Issue #2, opened 2015-10-22
Add support for getting files from S3


dancannon:
We are looking into using Nomad and store our binaries in S3. To fetch these as part of the job startup it would be helpful for go-getter to support S3.

I have started this [here](https://github.com/dancannon/go-getter/tree/s3) but I am not sure about my approach. I would be interested to hear your thoughts.

catsby:
Hey @dancannon – I imagine S3 is certainly a thing we'd like to support. Unless you feel it will be a lot of work, go ahead and get it working and then open a pull request and we'll be happy to review and provide suggestions! 

I did look at what you have and it seems straightforward enough. Do you have any ideas on how you plan on handling authentication? Defaulting to the aws-sdk-go to find them in ~/.aws/credentials or in the environment? That sounds sufficient for me, but maybe others have an opinion

dancannon:
Hey @catsby thats great to hear. Regarding my branch:
 
 - Providing the `Get` function might be a little difficult for S3 unless I have misunderstood what it is meant to be doing.
 - I thought that relying on aws-sdk-go for authentication would be enough but if not I imagine that credentials could be passed to the `S3Getter` to override the defaults, how does that sound?

dancannon:
Also what did you think about my method for matching S3 URLs? My current attempt will cause any S3 urls to use the S3 getter and not the HTTP getter, I dont think this is a bad thing but what do you think?

mitchellh:
Hey @dancannon. What you're starting looks great! I've put some points below in no particular order to help you out:

* All of go-getter is unit tested quite well. Take a look at how we test everything and try to copy! In particular, use table tests for detectors so we can see how you're detecting URLs.
* For auth, I think we should use all the defaults for the aws-sdk-go SDK but also probably allow for them from URL query params (that we parse out). 
* For `Get`, we should probably list the prefix and download all the files. I think S3 can do this quite well.
* Please write tests even for the Get methods using some publicly available S3 URL. We can replace this with our own S3 URL on merge so your account isn't on the line!

Let me know if you have any other questions and we'd _love_ to have this in.

mitchellh:
Done, by you!

dancannon:
Thanks for reviewing + merging!

PR #3, opened 2015-10-23
Add support for S3

dancannon:
As discussed in #2 I have worked on adding support for S3, I have also included the features requested in the mentioned ticket, to summarise this adds the following features:

 - Detect S3 URLS in the following formats (this means that any S3 urls will no longer be fetched using the HTTP getter)
  - `BUCKET.s3.amazonaws.com/PATH`
  - `BUCKET.s3-REGION.amazonaws.com/PATH`
  - `s3.amazonaws.com/BUCKET/PATH`
  - `s3-REGION.amazonaws.com/BUCKET/PATH`

- Also supported is the ability to override the default authentication method and the version being "getted" (this is only supported by `GetFile`). For example:
 - `s3-REGION.amazonaws.com/BUCKET/PATH?aws_access_key_id=foo&aws_access_key_secret=bar&aws_access_token=baz`
 - `s3-REGION.amazonaws.com/BUCKET/PATH?version=1234`

- Using `Get` all files with the given prefix are fetched and stored in the destination path relative to the given prefix
- Using `GetFile` fetches the given file from S3 and is stored at the destination path. As mentioned it is also possible to fetch a versioned S3 object. 

If you have any further questions/feature suggestions let me know, thanks.

### DIFF
--- a/detect_s3.go
+++ b/detect_s3.go
@@ -28,7 +28,7 @@ func (d *S3Detector) Detect(src, _ string) (string, bool, error) {
 
 func (d *S3Detector) detectHTTP(src string) (string, bool, error) {
 	parts := strings.Split(src, "/")
-	if len(parts) < 0 {
+	if len(parts) < 2 {
 		return "", false, fmt.Errorf(
 			"URL is not a valid S3 URL")
 	}
@@ -48,7 +48,7 @@ func (d *S3Detector) detectPathStyle(region string, parts []string) (string, boo
 	urlStr := fmt.Sprintf("https://%s.amazonaws.com/%s", region, strings.Join(parts, "/"))
 	url, err := url.Parse(urlStr)
 	if err != nil {
-		return "", true, fmt.Errorf("error parsing GitHub URL: %s", err)
+		return "", false, fmt.Errorf("error parsing S3 URL: %s", err)
 	}
 
 	return "s3::" + url.String(), true, nil
@@ -58,7 +58,7 @@ func (d *S3Detector) detectVhostStyle(region, bucket string, parts []string) (st
 	urlStr := fmt.Sprintf("https://%s.amazonaws.com/%s/%s", region, bucket, strings.Join(parts, "/"))
 	url, err := url.Parse(urlStr)
 	if err != nil {
-		return "", true, fmt.Errorf("error parsing S3 URL: %s", err)
+		return "", false, fmt.Errorf("error parsing S3 URL: %s", err)
 	}
 
 	return "s3::" + url.String(), true, nil
--- a/detect_s3.go
+++ b/detect_s3.go
@@ -1,0 +1,89 @@
+package getter
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+const (
+	vhostFormat = ""
+)
+
+// S3Detector implements Detector to detect S3 URLs and turn
+// them into URLs that the S3 getter can understand.
+type S3Detector struct{}
+
+func (d *S3Detector) Detect(src, _ string) (string, bool, error) {
+	if len(src) == 0 {
+		return "", false, nil
+	}
+
+	if strings.Contains(src, ".amazonaws.com/") {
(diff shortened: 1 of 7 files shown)


detect_s3.go
```diff
@@ -0,0 +1,89 @@
...
+	urlStr := fmt.Sprintf("https://%s.amazonaws.com/%s/%s", region, bucket, strings.Join(parts, "/"))
+	url, err := url.Parse(urlStr)
+	if err != nil {
+		return "", true, fmt.Errorf("error parsing S3 URL: %s", err)
+	}
+
+	return "s3::" + url.String(), true, nil
+}
+
+// func (d *S3Detector) detectSSH(src string) (string, bool, error) {
```
catsby: looks like this was left over from a copy-paste

dancannon: Removing now, thanks.

catsby:
Hey @dancannon – great start!
When I add my credentials to my environment, I get a single failure:

```
--- FAIL: TestS3Getter (0.00s)
	get_s3_test.go:21: err: URL is not a valid S3 URL
```

Is that expected, because I don't have access to that URL? It _sounds_ like an invalid URL, not just a permissions thing

dancannon:
It sounds like something else is going on, could you please paste a sample URL that has the same format?

catsby:
This is the test and url that's failing 

- https://github.com/hashicorp/go-getter/pull/3/files#diff-fe4fbe1b573926567435619dee064738R20

The URL is `https://s3-eu-west-1.amazonaws.com/hailo-s3-test`

cUrl suggest the url itself is fine:

```
$ curl -I https://s3-eu-west-1.amazonaws.com/hailo-s3-test
HTTP/1.1 200 OK
```

dancannon:
Ah sorry about that!! I had added some more validation which wasnt quite correct. The tests should pass now.

detect_s3.go
```diff
@@ -0,0 +1,65 @@
+package getter
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+const (
+	vhostFormat = ""
```
mitchellh: Unused constant, remove?

detect_s3_test.go
```diff
@@ -0,0 +1,84 @@
...
+		},
+		{
+			"s3-eu-west-1.amazonaws.com/bucket/foo/bar.baz",
+			"s3::https://s3-eu-west-1.amazonaws.com/bucket/foo/bar.baz",
+		},
+		// Misc tests
+		{
+			"s3-eu-west-1.amazonaws.com/bucket/foo/bar.baz?version=1234",
+			"s3::https://s3-eu-west-1.amazonaws.com/bucket/foo/bar.baz?version=1234",
+		},
```
mitchellh: These tests look fantastic. Great job!

mitchellh:
The only thing that appears to be missing is loading credentials from environment, profile, etc. The AWS SDK does this for you but you have to make sure to just enable the credential lookup thing. If you already do this and I missed it or you need more guidance, let me know.

Your tests look great.

dancannon:
Thanks for looking at this so quickly, regarding the AWS credentials I was under the impression that if you dont pass any Credentials to the config then [DefaultChainCredentials](https://github.com/aws/aws-sdk-go/blob/66c840e9981dd121a4239fc25e33b6c1c1caa781/aws/defaults/defaults.go#L18) is used by default which looks up the credentials. Also I tested this code with a private S3 bucket and it seemed to pick up the credentials.

If this is incorrect let me know and I will look into it again.

mitchellh:
We should explicitly setup the credential chain. By using `DefaultChainCredentials` it allows an external caller to modify it and we don't really want to allow that. Once that fix is in, I'll merge

dancannon:
I have made the suggested change, however now that you had added CI the tests are failing as travis does not have any AWS credentials.

Have you got any ideas for how we could fix this?

mitchellh:
I'll take a look locally. Thanks!

PR #4, opened 2015-10-24
Update git branches properly

mitchellh:
Fixes https://github.com/hashicorp/terraform/issues/1568

Branches were not updated properly before. This fixes that by checking if we're on a branch and running `git pull` on the proper branch.


mitchellh:
Verbal LGTM!

PR #5, opened 2015-10-25
Run Detect automatically on Client

mitchellh:
/cc @catsby 

This modifies the Client to support running Detect automatically. Tests and docs are in.


catsby:
LGTM!

PR #6, opened 2015-10-27
Break cyclic dependecny on terraform

avnik:


mitchellh:
I just ran `make dev` on Terraform and there was no cycle. We import from each other but it is acyclic so it works fine. At some point we'll probably extract that library but not yet. Thanks though!

PR #8, opened 2015-11-13
Move url helper to go-getter from terraform

radeksimko:
There's currently a circular dependency between `hashicorp/terraform` and `hashicorp/go-getter` caused by this package and the package is only being used in `go-getter` (based on quick `grep` & github search) so I don't see a reason why keep it in terraform.

After this is merged, another PR in terraform for removal of `helper/url` directory can be raised.

It's probably not the best solution for https://github.com/hashicorp/terraform/pull/3773 , but I think it's good to avoid circular dependencies anyway, where possible.

### DIFF
--- a/client.go
+++ b/client.go
@@ -15,7 +15,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	urlhelper "github.com/hashicorp/terraform/helper/url"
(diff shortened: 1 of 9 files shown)


mitchellh:
Ah, someone brought this up before but I didn't realize this is only used by go-getter. In that case let's do it.

PR #9, opened 2016-01-12
Unarchiving Support

mitchellh:
This adds support for downloading archives with go-getter and automatically unarchiving them.

We use the magic query parameter `archive` to specify the archive type. We also have shorthand by detecting the extension and using that to populate `archive` if it isn't set. Supported archive types are: `tar.gz`, `tar.bz2`, `gz`, `bz2`, `tgz`, `tbz`, and `zip`. More types can be very easily added.

Archiving can be disabled explicitly (if you're actually trying to just download the archive itself) by setting `archive=false` (or 0, anything that ParseBool's to false). 

This will instantly enable the following in our projects:

  * Terraform modules can be archived
  * Nomad binaries can be archived
  * Otto dependencies and imports can be archived


client.go
```diff
@@ -105,10 +119,59 @@ func (c *Client) Get() error {
 			"download not supported for scheme '%s'", force)
 	}
 
+	// We have magic query parameters that we use to signal different features
+	q := u.Query()
+
+	// Determine if we have an archive type
+	archiveV := q.Get("archive")
+	if archiveV != "" {
+		q.Del("archive")
```
dadgar: I would put a comment on why you delete

client.go
```diff
@@ -211,3 +300,12 @@ func checksum(source string, h hash.Hash, v []byte) error {
 
 	return nil
 }
+
+var archiveMagic = map[string]struct{}{
```
dadgar: Unused?

mitchellh: Fixed in a later commit :)

PR #10, opened 2016-02-25
GetAny

mitchellh:
`GetAny` will attempt to download a file or directory. The semantics are:

  * If the URL is a file, it downloads it into `dst` with the basename. For example, downloading "/path/foo.txt" will store it in "dst/foo.txt"

  * If the URL is an archive, it will be unpacked into `dst`.

Directories aren't currently supported but will be.

### DIFF
--- a/client.go
+++ b/client.go
@@ -222,6 +222,15 @@ func (c *Client) Get() error {
 		checksumValue = b
 	}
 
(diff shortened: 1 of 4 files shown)


dadgar:
Looks great!

mitchellh:
Lets do this

PR #12, opened 2016-03-20
Remove reference to hashicorp/terraform/config

jen20:
This commit removes a reference to the hashicorp/terraform/config package which was unused but causing builds to complain about it being unavailable.


phinze:
LGTM, kill all dead code. :skull: :gun: :smile:

PR #14, opened 2016-03-20
hg/windows: Fix GetFile

jen20:
This commit fixes the TestHgGetter_GetFile test which was failing on Windows because of a missing call to filepath.Slash and failing to set the host of the file URL to localhost, which is required for Mercurial on Windows.

NOTE: this has not fixed all of the red tests on Windows so AppVeyor failing is still expected. Prior to merging it should be verified that the "TestHgGetter_GetFile" test is passing on Windows and that Travis is passing.


phinze:
This does indeed fix `TestHgGetter_GetFile` - LGTM!

PR #15, opened 2016-03-20
tgz/windows: Fix the TestTarGzipDecompressor test

jen20:
The expected format for Windows uses backslashes instead of forward slashes.

This _should_ get AppVeyor green!


phinze:
Green green green! LGTM

PR #16, opened 2016-03-20
Use junction points instead of symlinks on Windows

jen20:
This resolves a problem observed on domain joined machines where the user has no Administrative priviledges, which is a prerequisite for creating a symbolic link in Windows. This breaks local Terraform modules on such machines.

Instead for directories we shell to `mklink /J` which creates a junction point - this is inspired by the Go standard library test suite.


phinze:
LGTM

PR #17, opened 2016-03-21
decompressor/zip: fix behavior with subdirectories

### LINKED ISSUE #11: Zip files that contain directories that don't exist on dest fail
If a zip file contains: `new_folder/new_file` and I tell go-getter to download it to `dest/` which doesn't contain `new_folder` I get the following error:

open /tmp/NomadClient446359139/1491681f-974e-9e16-3276-7d3ac7aa5b1a/redis/local/bin/cp: no such file or directory

mitchellh:
Fixes #11 

This fixes two bugs:

  * Empty directories had an incorrect boolean statement causing them to not unzip.

  * Zips that contained files in subdirectories without an empty subdirectory entry would fail to decompress.

Test cases for all of this are added.


decompress_zip.go
```diff
@@ -46,7 +46,7 @@ func (d *ZipDecompressor) Decompress(dst, src string, dir bool) error {
 		}
 
 		if f.FileInfo().IsDir() {
-			if dir {
+			if !dir {
```
dadgar: I think this is how we were seeing this error [here](https://github.com/hashicorp/nomad/issues/923)

dadgar:
LGTM

decompress_zip.go
mitchellh: That is right. This will fix that.

PR #18, opened 2016-03-23
Client tests detectors

testbetta1:


mitchellh:
I don't think you intended to open this, perhaps.

PR #29, opened 2016-08-20
Netrc support for HTTP requests

mitchellh:
This adds automatic [netrc](http://linux.die.net/man/5/netrc) support for HTTP URLs. 

Netrc is generally not recommended since auth information is stored in plaintext however it is still very much in use and in a pinch is useful to provide auth information for something you don't want to exist in the source itself. 

This PR does the following:

  * If no username is specified for an HTTP URL, netrc is checked for a matching host. If found, the auth info is automatically used

  * The host in the netrc file must be an exact match or the "default" host

  * The `NETRC` env var is respected as an alternate path 

  * Path for netrc is `$HOME/.netrc` on Unix-like systems or `%HOME%/_netrc` on Windows

**NOTE ON GIT SUPPORT:** `git` uses `curl` under the covers which automatically uses netrc (and follows the same rules above) for HTTP URLs. Therefore, go-getter doesn't need to do anything there.

### DIFF
--- a/get.go
+++ b/get.go
@@ -46,7 +46,7 @@ var Getters map[string]Getter
 var forcedRegexp = regexp.MustCompile(`^([A-Za-z0-9]+)::(.+)$`)
 
 func init() {
-	httpGetter := new(HttpGetter)
+	httpGetter := &HttpGetter{Netrc: true}
 
 	Getters = map[string]Getter{
 		"file":  new(FileGetter),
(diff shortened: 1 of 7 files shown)


mitchellh:
Failure in CI is unrelated to this PR and I'll try to fix in a diff commit/PR.

jen20:
LGTM.

mitchellh:
Get excited.

Ashald:
Does it support https requests?

PR #30, opened 2016-08-24
Use url.RawPath for paths with a %2F in them

mitchellh:
https://github.com/hashicorp/terraform/issues/8418

If a URL encoded element is in a file path, it was not working. It appears that `url.Path` doesn't include it but `url.RawPath` does include it. I'm not 100% confident on the semantics of each despite reading the docs and source... but this keeps all tests passing.


mitchellh:
CI failures are unrelated to this PR.

phinze:
LGTM

PR #31, opened 2016-09-12
Resolve symlinked PWD in FileDetector.Detect

### LINKED ISSUE #28: File getter does not work with relative paths from a relative symlinked directory
Getting a file with a local path (say `./filename`) returns an error if you are running `go-getter` or `terraform` from a symlinked directory.

Reproducer:

```bash
# this works
cd src/github.com/hashicorp/go-getter
go-getter -mode file ./module_test.go ./foo
2016/07/13 10:40:46 Success!

# this does not work
cd ..
ln -sf go-getter go-getter2
cd go-getter2/
go-getter -mode file ./module_test.go ./foo
2016/07/13 10:41:31 Error downloading: source path error: stat /go-getter/module_test.go: no such file or directory
```

Update: works as expected if the symlink is a full, non-symlinked file.

jbardin:
When a relative symlink is provided for the pwd in FileDetector.Detect,
it can result in an invalid URL relative to the file root, e.g.
`file:///../modules/foo`. Resolve the absolute path after reading a
symlink in Detect.

Fixes #28

### DIFF
--- a/detect_file.go
+++ b/detect_file.go
@@ -36,6 +36,13 @@ func (d *FileDetector) Detect(src, pwd string) (string, bool, error) {
 				if err != nil {
 					return "", true, err
 				}
+
+				// The symlink itself might be a relative path, so we have to
+				// resolve this to have a correctly rooted URL.
+				pwd, err = filepath.Abs(pwd)
+				if err != nil {
+					return "", true, err
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


mitchellh:
LGTM! 

Symlinks are just the worst. :)

jbardin:
I don't have commit access here, but I htink this is ready to merge

PR #37, opened 2016-10-20
decompressor_tgz: handle symlinks and links (hardlinks)

pearkes:
This adds support for symlinks and hardlinks being unpacked from `tar.gz` archives go-getter processes. Previously, these files would be copied from their empty contents from the target `tar.gz` to the destination file system.

I have a few questions:
1. Is the right approach for hard links? I assume it wouldn't be possible due to the target needing to exist prior to linking – curious if there are good patterns to deal with that.
2. How do we do this in other compressors?
3. The table driven tests only inspect file contents for comparison (necessary for ensuring that this functionality works as intended) with single files – should I just augment the `TestDecompressCase` test case to accept multiple MD5s for comparison, or maybe just write an individual test case outside of the table driven style to test this?


schmichael:
:wave: 

I started playing around with symlink, hardlink, and device node (:grimacing:) support in a fork because an approach we were taking for LXC in nomad required it. [`branch`](https://github.com/hashicorp/go-getter/compare/master...schmichael:f-tar-devices-links) We ended up not needing it though, so I just pushed the unfinished go-getter branch for posterity.

At first I was afraid supporting links posed a security issue for nomad\* -- however it's no different from the existing security issue of absolute paths in archives.

So! Here's my thinking:

1. Ship this once you fix @ryanuber's comments
2. Treat tars being able to extract files anywhere on the filesystem as a separate issue/PR to be addressed as needed (same with device nodes which I think require a root user or special capability and will probably only ever be needed in nomad)
3. Either in this PR or another refactor _tar_ handling into a shared function that's given an `io.Reader` wrapped in the appropriate decompressor.

<sup>\*</sup> eg tar contains symlink `foo` to `/etc` and then extracts `foo/passwd`

langmartin:
Closing because this pr has been replaced by #192

PR #38, opened 2016-10-27
FIX: Private enterprise github references

johnthedev97:
I was redirected from here https://github.com/hashicorp/terraform/pull/9644

Added a few unit test cases for github detector
Removed the hardcoded domain ‘github.com’ in detectSSH() and set it from
the source
Modified the initial testing logic for HTTPS in Detect() such that
handles domain names where github is a subdomain

PS: The test cases was added by looking at the code and behavior before
my code changes and not from knowledge of use cases/expected values. My
main objective for the test cases was to ensure I don’t break the
behavior regular github.com URLs

### DIFF
--- a/detect_github.go
+++ b/detect_github.go
@@ -15,9 +15,11 @@ func (d *GitHubDetector) Detect(src, _ string) (string, bool, error) {
 		return "", false, nil
 	}
 
-	if strings.HasPrefix(src, "github.com/") {
+	parts := strings.Split(strings.Split(src, "/")[0], ".")
+
+	if parts[0] == "github" && parts[len(parts)-1] == "com" {
 		return d.detectHTTP(src)
-	} else if strings.HasPrefix(src, "git@github.com:") {
+	} else if strings.HasPrefix(src, "git@") {
 		return d.detectSSH(src)
 	}
 
@@ -51,6 +53,7 @@ func (d *GitHubDetector) detectHTTP(src string) (string, bool, error) {
 func (d *GitHubDetector) detectSSH(src string) (string, bool, error) {
 	idx := strings.Index(src, ":")
 	qidx := strings.Index(src, "?")
+	hidx := strings.Index(src, "@")
 	if qidx == -1 {
 		qidx = len(src)
 	}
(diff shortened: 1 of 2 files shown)


mitchellh:
This looks good, but not being knowledgeable about this is it a fair assumption that if `github` is in the domain, that we are talking to a private GitHub instance? I suppose thats more likely than not but I wonder if that is too weird...

johnthedev97:
I cannot say for sure, but if github is sub domain (github.company.com) I thought it might be fair assumption. It was in my case (I faced issue accessing my company's private github in terraform module source over ssh)

johnthedev97:
Let me know if you have any suggestions on ascertaining the URL refers to github.

I think for SSH, as the user explicitly start with 'git@' it is fine. I mean even if it is not github and is some other git provider it would work. 

If you want I can remove the changes for HTTP, as that is more subjective

moofish32:
@mitchellh - I have the same github enterprise configuration at my company and the same feature request. I could agree on only opening up ssh access, based on `git@` and not the http(s) interface.  Is there any chance on this making it into master or is the lack of a robust detection mechanism a blocker?

rbhitchcock:
I ran into this issue today while trying to import a Terraform module from my company's Github Enterprise install.

```hcl
module "foo" {
  source = "git::git@github.company.com/org/moduless.git//dag?ref=mybranch"
}
```

Based on the existing behavior, it treats the source as a local `file://` URI as opposed to a git URI. Is there anything I can do to help this change be merged? I'm happy to make any modifications to meet your requirements @mitchellh (assuming @johnthedev97 has moved on to other things and no longer has time for this).

My 2 cents would be to back out the changes for HTTP detection. A workaround for this already exists, e.g. `source = git::https://github.company.com/org/modules.git`. The changes for SSH detection by looking for `git@` instead of `git@github` seem to be the most important.

Please let me know if there's anything I can do to help this along!

johnthedev97:
@rbhitchcock if it helps, I went ahead with accessing github via SSH using the format - "git::ssh://git@github.company.com/...." which works without this patch

rbhitchcock:
@johnthedev97 Thank you!

mitchellh:
This should be fixed by #129 very soon. 

I didn't do the magic behavior where if "github" and "com" are in the URL I assume GitHub. However, SCP-like URLs like `git@anyhost.com:path` will now work as you expect! This should make all the GHE access work.

PR #40, opened 2016-12-05
Bug/false decompressor detect

Nomon:
Fix an issue that prevents getting folders that have the same name as a decompressor.

Example terraform file that fails with error "source path must be a file":
```
module "artifact" {
  source = "../artifacts/zip"
}
```

The change changes the matching to include a dot (ie .zip) which makes much more sense for auto detection.


Nomon:
Closing PR due to https://github.com/hashicorp/go-getter/pull/49

Issue #25, opened 2016-04-29
Broken packge state?


vtolstov:
```
go-getter git::https://github.com/yoctocloud/packer-centos.git ./tests
2016/04/29 18:39:01 Error downloading: /usr/host/bin/git exited with 128: Cloning into '/tmp/getter-git175513477'...
fatal: repository 'https://github.com/yoctocloud/' not found
```

why command strips last component?
i'm try use go-getter lib in my app and seems that it strips last component too.

vtolstov:
As i see mode Any thinks in this case that this is file packer-centos with ext git.
Why this happened ? I'm specify protocol git....

ryanuber:
@vtolstov I think this should be resolved after #41 above, so I'm going to close this. Let me know if you get a chance to try it out or if you still see this issue when building from master. Thanks!

PR #41, opened 2017-01-05
Implement GetAny()

### LINKED ISSUE #35: Running go-getter always returns 128 for a github repo.
The command go-getter github.com/hashicorp/nomad ./test returns the error

2016/10/05 10:29:50 Error downloading: /usr/bin/git exited with 128: Cloning into '/tmp/getter-git236954800'...
fatal: repository 'https://github.com/hashicorp/' not found

Am I misunderstanding? Or can someone help

ryanuber:
Implements the getter-selected client mode when `ClientModeAny` or `GetAny()` are used. This comes with the caveat that clients may be depending on the currently-implemented `ClientModeAny`, which simply assumes everything is a file. This behavior seems broken anyways, so hopefully this isn't a big deal.

/cc @mitchellh

Fixes #35, #25

### DIFF
--- a/get_file.go
+++ b/get_file.go
@@ -1,6 +1,9 @@
 package getter
 
-import "net/url"
+import (
+	"net/url"
+	"os"
+)
 
 // FileGetter is a Getter implementation that will download a module from
 // a file scheme.
@@ -9,6 +12,15 @@ type FileGetter struct {
 	Copy bool
 }
 
-func (g *FileGetter) ClientMode(_ *url.URL) ClientMode {
+func (g *FileGetter) ClientMode(u *url.URL) ClientMode {
+	path := u.Path
+	if u.RawPath != "" {
+		path = u.RawPath
+	}
+
+	if fi, err := os.Stat(path); err == nil && fi.IsDir() {
+		return ClientModeDir
+	}
+
 	return ClientModeFile
 }
--- a/get_file.go
+++ b/get_file.go
@@ -12,15 +12,15 @@ type FileGetter struct {
 	Copy bool
 }
 
-func (g *FileGetter) ClientMode(u *url.URL) ClientMode {
+func (g *FileGetter) ClientMode(u *url.URL) (ClientMode, error) {
 	path := u.Path
 	if u.RawPath != "" {
 		path = u.RawPath
 	}
 
 	if fi, err := os.Stat(path); err == nil && fi.IsDir() {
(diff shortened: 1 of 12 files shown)


get_git.go
```diff
@@ -15,6 +15,10 @@ import (
 // a git repository.
 type GitGetter struct{}
 
+func (g *GitGetter) ClientMode(_ *url.URL) (ClientMode, error) {
+	return ClientModeDir, nil
```
ryanuber: This seemed the best we could do for the 95% case in the VCS getters. Although we do have `GetFile()` for the git getter, it is not usable because of the way the source and dest are parsed out of the URL. It seems like it will take some thought to get it right in a generic way for `git` and `hg`.

evanphx:
Looks good. Trying to auto detect the mode is difficult and I think this does the best possible job.

get_file.go
```diff
@@ -1,8 +1,27 @@
...
 }
+
+func (g *FileGetter) ClientMode(u *url.URL) (ClientMode, error) {
+	path := u.Path
+	if u.RawPath != "" {
+		path = u.RawPath
+	}
+
+	// Check if the source is a directory.
+	if fi, err := os.Stat(path); err == nil && fi.IsDir() {
```
cbednarski: Should the stat check be extended to whether or not the file exists?

E.g. it looks like the logic says "if I can't stat return ClientModeFile" but this should be an error?

Otherwise seems odd because the validation is asymmetrical. I guess it could be assumed to be checked elsewhere but not clear from this method context.

get_mock.go
```diff
@@ -43,3 +43,10 @@ func (g *MockGetter) GetFile(dst string, u *url.URL) error {
 	}
 	return g.GetFileErr
 }
+
+func (g *MockGetter) ClientMode(u *url.URL) (ClientMode, error) {
+	if u.Path[len(u.Path)-1:] == "/" {
```
cbednarski: Maybe don't need comprehensive error checking in a mock, but it looks like this will panic if the URL is `""`.

get_file.go
ryanuber: Ah, good catch. I originally didn't have a way of returning an error from `ClientMode()`, but that changed later and I didn't update here.

get_git.go
mitchellh: Seems fine for now.

PR #43, opened 2017-01-20
Support providing SSH keys during git clones

ryanuber:
Adds an `?sshkey` query parameter which can be used to supply a private key during git clones over SSH. Use of this feature obviously requires some scrutiny. I think this is just enough to be useful while matching the general flow we have already.


cbednarski:
LGTM

PR #44, opened 2017-01-20
Clone git submodules recursively

ryanuber:
Automatically clones git submodules recursively.


mitchellh:
Looks like you have some test failures. Once you clean those up it LGTM.

ryanuber:
Weird! It looks like relative submodule URL's are handled a little differently between my local git version and what Travis is running. To fix that, I replaced the git-related fixtures with a test helper to create temporary git repos.

PR #48, opened 2017-02-07
Git test should use "echo" and trim space locally

mitchellh:
On macOS this test was failing for me since "echo" for /bin/sh doesn't
support "-n". I think the easiest solution is to just do that behavior
locally.


ryanuber:
LGTM :ship:

PR #49, opened 2017-02-07
Decompressors should check for "." in extension

mitchellh:
Fixes https://github.com/hashicorp/terraform/issues/11438

Decompressors were suffix checking only for the extension WITHOUT a
period. This was causing directories and files ending with that to be
treated as an archive incorrectly.


ryanuber:
LGTM! :ship:

PR #51, opened 2017-02-20
 Windows hardlinks should be symlinks

matthewhartstonge:
Hard links (NTFS junction point) should be soft link (symlinks) to match unix behaviour. 
See: hashicorp/terraform#11898


jbardin:
I'm not certain about Windows behavior here, but I think this may re-introduce the issue solved by #16. I'll look into a different solution too.

PR #53, opened 2017-03-08
Handle subdirs in tar.bz2 archives.

c2nes:
This fix should match the same fix applied for tar.gz files in
cd92db7af plus the follow on test fix in cf14e661.


mitchellh:
Fixed via #57. Sorry for not seeing yours first :( Thank you for the contribution.

PR #54, opened 2017-03-08
Handle missing subdirs

c2nes:
Handle missing directory entries in tar archives.

Some tools (at least Maven's assembly plugin) can produce tar
archives with missing directory entries. These archives will have files
in subdirectories without entries for the subdirectories themselves.

The fix implemented here is to always ensure the parent directory of a
file before opening the file itself.

The test archives included here have a single file "file1" within a
subdirectory "dir" (note the missing "dir/" entry),

    $ tar tf test-fixtures/decompress-tgz/missing_dir.tar.gz
    dir/file1

This PR is actually for the second commit here (2e149a8), but depends on the first. I've submitted #53 for the first commit.


c2nes:
I originally hit this issue through Nomad which was running into a "no such file or directory" issue while attempting to unpack an artifact matching this description.

mitchellh:
Sorry #57 fixed this. I should've seen this over 20 days ago. My mistake, thank you for the contribution.

winduptoy:
@c2nes I ran into the same thing. Is there any way to craft a tar that lists directories to work around this issue until it is merged into Nomad 0.6.0?

c2nes:
@winduptoy GNU tar and the version of tar included on OS X both included the directories for me. I only experienced this issue with archives generated by the Maven assembly plugin.

winduptoy:
@c2nes Using GNU tar, it seems that when I pass a directory, it works:

```
$ tar -czvf out.tar.gz dir1
dir1/
dir1/fileA
dir1/fileB
...
```

but when I pass a file within a subdirectory, I run into this issue:

```
$ tar -czvf out.tar.gz dir1/fileA
dir1/fileA
```
(no `dir1/` entry)

How lame. Here's my hack:

```
$ tar -czvf out.tar.gz --no-recursion dir1 --recursion dir1/fileA
```

PR #57, opened 2017-04-04
Remove assumption about file ordering in tars

dadgar:
This PR removes an assumption that a tar will list directories before
a file within that directory. There doesn't appear to be anything in the
tar specification regarding ordering so we should handle this.

Cross linking https://github.com/hashicorp/nomad/issues/2514


dadgar:
I also found the same bug here: https://github.com/hashicorp/go-getter/pull/53
So that one can be closed as I fixed it.

mitchellh:
LGTM. Thanks!

PR #58, opened 2017-04-05
Fix missing NETRC in get_http/GetFile

Ashald:
It was present only in `Get`, not in `GetFile`. And Nomad uses `GetFile` so no NETRC in Nomad so far.


dadgar:
LGTM thanks!

PR #59, opened 2017-04-30
Drop dependency on test from package

josephglanville:
Currently importing `go-getter` or any package that depends on it introduces a dependency on `testing`. If you happen to be using the inbuilt `flags` package this results in the `testing` flags polluting usage. It's also generally considered not to be good practice to depend on `testing` in non-test code.

### DIFF
--- a/decompress_bzip2_test.go
+++ b/decompress_bzip2_test.go
@@ -6,7 +6,7 @@ import (
 )
 
 func TestBzip2Decompressor(t *testing.T) {
-	cases := []TestDecompressCase{
+	cases := []decompressTestCase{
 		{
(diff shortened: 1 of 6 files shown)


mitchellh:
We do this on purpose so that consumers of the library can have a public and exported API for testing parts of the library. We do this in a number of our projects. However, we should be using https://github.com/mitchellh/go-testing-interface instead of "testing" so this doesn't happen. I'll fix that up now.

josephglanville:
Ahh I see, this solution works for me. 👍 

How long is this likely to take to bubble up into Terraform so that importing Terraform packages will benefit from this change?

PR #67, opened 2017-06-16
Use cleanhttp for the default http.Client

jbardin:
Since go-getter uses the `http.DefaultClient` for http requests, it's behavior may be inadvertently altered by other packages modifying the Client. 

Use `cleanhttp.DefaultClient` to get a new client instance.

### DIFF
--- a/get_http.go
+++ b/get_http.go
@@ -36,6 +36,10 @@ type HttpGetter struct {
 	// Netrc, if true, will lookup and use auth information found
 	// in the user's netrc file if available.
 	Netrc bool
+
+	// Client is the http.Client to use for Get requests.
+	// This defaults to http.DefaultClient if left unset.
+	Client *http.Client
 }
 
 func (g *HttpGetter) ClientMode(u *url.URL) (ClientMode, error) {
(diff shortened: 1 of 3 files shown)


get_http.go
```diff
@@ -57,13 +61,17 @@ func (g *HttpGetter) Get(dst string, u *url.URL) error {
 		}
 	}
 
+	if g.Client == nil {
+		g.Client = http.DefaultClient
```
ryanuber: This seems a little odd, what is the reason to default to the `http.DefaultClient` here? If it wasn't set it seems reasonable to use a fresh client from `cleanhttp`. Am I missing something?

get_http.go
```diff
@@ -106,7 +114,11 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
 		}
 	}
 
-	resp, err := http.Get(u.String())
+	if g.Client == nil {
+		g.Client = http.DefaultClient
```
ryanuber: Same as above, unsure why we want the `DefaultClient`in this case.

get_http.go
jbardin: This came from the first commit before cleanhttp, where the Client field was added, and it needed a default. I left it to preserve the existing behavior of a user making a new HTTPGetter and expecting it to inherit the `http.DefaultClient`, which is also documented on the struct field.

If you want to excise the old `http.DefaultClient` entirely, I can make a single global http client for this case, but we don't want to insert a new client every time we hit this because it will leak http connections.

ryanuber: Agreed it should be a private package global. IMO it's better to replace the client using the public `Getters` map on a `*getter.Client` if a custom client is needed. I think it's nicer to just use `cleanhttp` by default. We do the same in other packages like `go-retryablehttp`.

jbardin:
Pushed an update to remove `http.DefaultClient`

jbardin:
Sorry, missed the comment about a test.
Added one that verifies we no longer use `http.DefaultClient`.

PR #68, opened 2017-06-17
Add the ability to use minio with s3 getter

### LINKED ISSUE #52: S3: Custom url
Is it possible to use custom url? I'm looking to use minio with nomad artifacts

lfarnell:
This is an attempt at adding S3 compliant server support for the S3 protocol. I only tested with minio and it seemed to work

```bash

C:\go\src\github.com\hashicorp\go-getter\cmd\go-getter>go run main.go "s3::http://127.0.0.1:9000/test-bucket/hello.txt?aws_access_key_id=xxx&aws_access_key_secret=xxx" C:\
minio\

2017/06/17 01:37:26 Success!
```
I wasn't exactly sure how to add a unit test for it considering i would need to have a minio server running. Any feedback on the code would be great!

Closes #52

### DIFF
--- a/get_s3.go
+++ b/get_s3.go
@@ -28,7 +28,7 @@ func (g *S3Getter) ClientMode(u *url.URL) (ClientMode, error) {
 	}
 
 	// Create client config
-	config := g.getAWSConfig(region, creds)
+	config := g.getAWSConfig(region, u, creds)
 	sess := session.New(config)
 	client := s3.New(sess)
 
@@ -84,7 +84,7 @@ func (g *S3Getter) Get(dst string, u *url.URL) error {
 		return err
 	}
 
-	config := g.getAWSConfig(region, creds)
+	config := g.getAWSConfig(region, u, creds)
 	sess := session.New(config)
 	client := s3.New(sess)
 
@@ -139,7 +139,7 @@ func (g *S3Getter) GetFile(dst string, u *url.URL) error {
 		return err
 	}
 
-	config := g.getAWSConfig(region, creds)
+	config := g.getAWSConfig(region, u, creds)
 	sess := session.New(config)
 	client := s3.New(sess)
 	return g.getObject(client, dst, bucket, path, version)
@@ -174,7 +174,7 @@ func (g *S3Getter) getObject(client *s3.S3, dst, bucket, key, version string) er
 	return err
 }
 
-func (g *S3Getter) getAWSConfig(region string, creds *credentials.Credentials) *aws.Config {
+func (g *S3Getter) getAWSConfig(region string, url *url.URL, creds *credentials.Credentials) *aws.Config {
 	conf := &aws.Config{}
 	if creds == nil {
 		// Grab the metadata URL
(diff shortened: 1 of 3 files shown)


get_s3.go
```diff
@@ -204,29 +212,42 @@ func (g *S3Getter) getAWSConfig(region string, creds *credentials.Credentials) *
...
+	} else {
+		pathParts := strings.SplitN(u.Path, "/", 3)
+		if len(pathParts) != 3 {
+			err = fmt.Errorf("URL is not a valid S3 URL")
+			return
+		}
+		bucket = pathParts[1]
+		path = pathParts[2]
+		version = u.Query().Get("version")
+		region = "us-east-1"
```
dadgar: This should probably be parsable as well in case the user has set a non-default region on minio

get_s3.go
```diff
@@ -204,29 +212,42 @@ func (g *S3Getter) getAWSConfig(region string, creds *credentials.Credentials) *
...
 
 func (g *S3Getter) parseUrl(u *url.URL) (region, bucket, path, version string, creds *credentials.Credentials, err error) {
-	// Expected host style: s3.amazonaws.com. They always have 3 parts,
-	// although the first may differ if we're accessing a specific region.
-	hostParts := strings.Split(u.Host, ".")
-	if len(hostParts) != 3 {
-		err = fmt.Errorf("URL is not a valid S3 URL")
-		return
-	}
+	if strings.Contains(u.Host, "amazonaws.com") {
```
dadgar: Can you add a comment explaining the different parsing for supporting blob stores mimicking the AWS S3 API

get_s3.go
lfarnell: I completely overlooked the ability to set the region. I will query the region as well

get_s3.go
```diff
@@ -204,29 +212,45 @@ func (g *S3Getter) getAWSConfig(region string, creds *credentials.Credentials) *
 }
 
 func (g *S3Getter) parseUrl(u *url.URL) (region, bucket, path, version string, creds *credentials.Credentials, err error) {
```
dadgar: Lets add  some unit tests to test parsing both s3 and s3 complaint URLs. Lets us get some test coverage and avoid running the server in tests.

After that we can merge!

lfarnell:
@dadgar let me know if those unit tests are ok! Couldn't think of another way to test it

lfarnell:
Please dont merge yet. I want to clean up the tests a little and i forgot about a failure case. Sorry about this

lnguyen:
Question @armon are we looking to get this merge and added for 0.6.0 Nomad?

dadgar:
@lnguyen It would be great to get it in 0.6.0

@lfarnell Any update?

lfarnell:
@dadgar fixed my code and the tests. Ready for another review :+1:  Sorry for the delay!

get_s3_test.go
```diff
@@ -170,82 +169,77 @@ func TestS3Getter_ClientMode_collision(t *testing.T) {
...
+		bucket:  "test-bucket",
+		path:    "hello.txt",
+		version: "",
+	},
+}
 
-	if creds != nil {
-		t.Fatal("expected to not be nil")
+func TestS3Getter_Url(t *testing.T) {
+	for i, pt := range s3tests {
```
dadgar: I would just inline here:

```
var s3tests = []struct{
    url     string
  ...
  } {
      {
         // test 1
      },
      ...
  }
  for i, pt := range s3tests {
  }
   ````

And then you can use the new table driven test using t.Run(..)

Better example here: https://blog.golang.org/subtests

dadgar:
@lfarnell Looking good! Small comment on the tests just to clean it up a bit but otherwise 🥇

lfarnell:
@dadgar done!

PR #77, opened 2017-09-05
detector/file: pwd symlink should be determined with EvalSymlinks

mitchellh:
We were using `ReadLink` before which read the exact link.
Unfortunately, we want to evaluate this link relative to the pwd if it
is an absolute path otherwise we evaluate the symlink incorrectly.

For example, if the pwd is set to "/foo/fake" and the value is
"../real", but we're evaluating go-getter from the real shell pwd of
"/bar", then we'd turn "/foo/fake" into "/bar/real". The behavior of
shells is to take it relative to its location, not your pwd. This
preserves that.


dadgar:
LGTM. Probably should bump the travis build to 1.9

PR #80, opened 2017-09-12
don't override GIT_SSH_COMMAND from environment

jbardin:
The behavior of duplicate environment variables was previously
undefined, so in go1.9 the exec package was updated to specify that the
last instance of a variable wins, allowing for the expected behavior
when overriding a variable.

Rather than appending the GIT_SSH_COMMAND to the Command environment,
which would override previous settings, check the process environment
first, and append our options to any existing command.

### DIFF
--- a/get_git.go
+++ b/get_git.go
@@ -180,17 +180,34 @@ func (g *GitGetter) fetchSubmodules(dst, sshKeyFile string) error {
 // setupGitEnv sets up the environment for the given command. This is used to
 // pass configuration data to git and ssh and enables advanced cloning methods.
 func setupGitEnv(cmd *exec.Cmd, sshKeyFile string) {
-	var sshOpts []string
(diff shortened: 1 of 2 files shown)


get_git.go
```diff
@@ -180,17 +180,34 @@ func (g *GitGetter) fetchSubmodules(dst, sshKeyFile string) error {
 // setupGitEnv sets up the environment for the given command. This is used to
 // pass configuration data to git and ssh and enables advanced cloning methods.
 func setupGitEnv(cmd *exec.Cmd, sshKeyFile string) {
-	var sshOpts []string
+	gitSSHCommand := "GIT_SSH_COMMAND="
```
mitchellh: Nitpick I'd probably just do `const gitSSHCommand =` here since we never use this as a var. But it has no real functional effect here.

PR #89, opened 2017-11-09
Progress status

SwampDragons:
Creates a percent progress method that can be used to find out how far into a big download you are.  Didn't implement for git and Hg, since they don't have clients that readily return the information.

I did this because we want to use go-getter to handle downloads within Packer, but we need progress for the big isos that people grab.


SwampDragons:
Gonna take this a different direction.

PR #90, opened 2017-11-22
Ranged requests

SwampDragons:
Add capability to handle ranged http requests.  based heavily off Packer's current download code.

**TODO before merge** (based on PR feedback below):
- [x] implement actual download tests using ranged requests and a dummy file
- [x] Clean up logic and error related to bad ranged request resulting in full file download
- [x] Add comment clarifying why we're allocating http.NewRequest up front
- [x] move error check for rangeErr earlier and make sure to address edge case mentioned https://github.com/hashicorp/go-getter/pull/90/- files/4c10796653a5da59662524ba3c9d7d261af4dc61#r157674731
- [x] remove unnecessary parens on l 145
- [x] log when HEAD request fails -- If we intend to ignore errors and download the full file, we should probably at least log that the HEAD - request failed so range request will be ignored.
- [x] validate whether if resp.StatusCode != 200 && !(isPartialDownload && resp.StatusCode == 206)` will work on line 166
- [x] don't export getByteRange (lowercase it)
- [x] document getByteRange with comment (see https://github.com/hashicorp/go-getter/pull/90/- files/4c10796653a5da59662524ba3c9d7d261af4dc61#r157673072)
- [x] remove bytes_range_retval and just return nil, nil or nil, error.
- [x] make errMsg a package scoped constant at bottom of file
- [x] make sure that errMsg doesn't obscure actual error by wrapping a new error
- [x] turn tests into a table test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,6 +256,12 @@ None
   * `rev` - The Mercurial revision to checkout.
 
 ### HTTP (`http`)
+  * `ranged_request_bytes` - If the server supports range requests, this will
+  set the `Range` header when making the HTTP request, allowing you to specify
+  a specific range of bytes to download.
+  Example usage: `http://www.path/2/my/file.iso?ranged_request_bytes=5555-66666`,
+  or `http://www.path/2/my/file.iso?ranged_request_bytes=5555-` which will
+  request byte 5555 to EOF.
 
 #### Basic Authentication
 
--- a/get_http.go
+++ b/get_http.go
@@ -4,10 +4,12 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/hashicorp/go-safetemp"
@@ -131,19 +133,60 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
 		g.Client = httpClient
 	}
 
+	// check to see whether user has specified a range of bytes to download.
+	// if user has, but the range is invalid, fall back to downloading the
+	// whole file
+	byteRange, rangeErr := getByteRange(u)
+	if rangeErr != nil {
+		// log that we are downloading whole file even though user
+		// requested a byte range.
+		log.Printf("%s; going to disregard range request and download entire file",
+			rangeErr)
+	}
+
+	// First determine whether we can make a ranged request. Then download.
 	req, err := http.NewRequest("GET", u.String(), nil)
 	if err != nil {
 		return err
 	}
 
-	req.Header = g.Header
+	isPartialDownload := false
+	if rangeErr == nil && byteRange != nil {
+		// We first make a HEAD request so we can check if the server supports
+		// range queries. If the server/URL doesn't support HEAD requests,
+		// we just fall back to GET.
+		resp, err := g.Client.Head(u.String())
+		if err == nil {
+			defer resp.Body.Close()
+		}
+
+		if err == nil && resp.StatusCode >= 200 && resp.StatusCode < 300 {
+			// If the HEAD request succeeded, then attempt to set the range
+			// query if we can.
+			if resp.Header.Get("Accept-Ranges") == "bytes" {
+				req.Header.Set("Range", fmt.Sprintf("bytes=%s-%s",
+					byteRange[0], byteRange[1]))
+				isPartialDownload = true
+			}
+		} else {
+			log.Printf("HEAD request for Range failed; falling back to full file download: %s",
+				err)
+		}
+	}
+
 	resp, err := g.Client.Do(req)
 	if err != nil {
 		return err
 	}
 
+	req.Header = g.Header
+	if err != nil {
+		return err
+	}
+
 	defer resp.Body.Close()
-	if resp.StatusCode != 200 {
+	// If we did a ranged request we should get a 206; otherwise we should get a 200
+	if resp.StatusCode != 200 && !(isPartialDownload && resp.StatusCode == 206) {
 		return fmt.Errorf("bad response code: %d", resp.StatusCode)
 	}
 
@@ -243,6 +286,62 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
 	}
 }
 
+// getByteRange is a helper function to parse out the byte range for ranged
+// requests.
+//
+// input values:
+// u must be non-nil and unmodified.
+// example of u:
+// "http://my/file.iso?ranged_request_bytes=5555-66666"
+// "http://my/file.iso?ranged_request_bytes=5555-"
+// note that the format of the raw url string can be either
+// bytes=startByte- OR bytes=startByte-endByte
+//
+// return values:
+// (non-nil, non-nil) - should not occur
+// (nil, nil) - returned when no byte range was provided in the first place
+// (nil, non-nil) - error occured when parsing or validating the byte range
+// (non-nil, nil) - byte range was successfully parsed, and the range is
+// returned as an array of strings in the following format:
+// example of a non-nil return value:
+// []string{"5555", "66666"}      // first url example above
+// []string{"5555", ""}             // second url example above
+func getByteRange(u *url.URL) ([]string, error) {
+	byteRange := u.Query().Get("ranged_request_bytes")
+	if byteRange == "" {
+		return nil, nil
+	}
+
+	vals := strings.Split(byteRange, "-")
+
+	// validate byte range values given
+	if len(vals) != 2 {
+		return nil, fmt.Errorf("%s; length of parsed byte range is not 2, %v",
+			invalidRangeMsg, vals)
+	}
+	start, err := strconv.ParseInt(vals[0], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("%s; could not convert start byte string to "+
+			"an integer: original error: %s", invalidRangeMsg, err)
+	}
+	// If you leave endBytes blank, read to end of file.
+	if vals[1] != "" {
+		// otherwise make sure "finish" value is a number
+		finish, err := strconv.ParseInt(vals[1], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("%s; could not convert finish byte string "+
+				"to an integer; original error: %s", invalidRangeMsg, err)
+		}
+		// need finish to be bigger than start val
+		if finish <= start {
+			return nil, fmt.Errorf("%s; finish byte must be bigger than start byte",
+				invalidRangeMsg)
+		}
+	}
+
+	return vals, nil
+}
+
 // attrValue returns the attribute value for the case-insensitive key
 // `name', or the empty string if nothing is found.
 func attrValue(attrs []xml.Attr, name string) string {
@@ -268,3 +367,5 @@ func charsetReader(charset string, input io.Reader) (io.Reader, error) {
 		return nil, fmt.Errorf("can't decode XML document using charset %q", charset)
 	}
 }
+
+const invalidRangeMsg = "Invalid byte range provided"
--- a/get_http_test.go
+++ b/get_http_test.go
@@ -8,7 +8,11 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
+	"time"
+
+	urlhelper "github.com/hashicorp/go-getter/helper/url"
 )
 
 func TestHttpGetter_impl(t *testing.T) {
@@ -187,6 +191,38 @@ func TestHttpGetter_file(t *testing.T) {
 	assertContents(t, dst, "Hello\n")
 }
 
+func TestHttpGetter_ranged_request(t *testing.T) {
+	var rangeReqs = []struct {
+		Input    string
+		Expected string
+	}{
+		{"0-", "Hello\n"},
+		{"0-2", "Hel"},
+		{"2-3", "ll"},
+		{"3-", "lo\n"},
+	}
+
+	ln := testHttpServer(t)
+	defer ln.Close()
+
+	g := new(HttpGetter)
+	dst := tempFile(t)
+	for _, rangeReq := range rangeReqs {
+		u, _ := urlhelper.Parse(fmt.Sprintf("http://%s/file?ranged_request_bytes=%s",
+			ln.Addr().String(), rangeReq.Input))
+		// Get it!
+		if err := g.GetFile(dst, u); err != nil {
+			t.Fatalf("err: %s", err)
+		}
+
+		// Verify the main file exists
+		if _, err := os.Stat(dst); err != nil {
+			t.Fatalf("err: %s", err)
+		}
+		assertContents(t, dst, rangeReq.Expected)
+	}
+}
+
 func TestHttpGetter_auth(t *testing.T) {
 	ln := testHttpServer(t)
 	defer ln.Close()
@@ -240,6 +276,48 @@ func TestHttpGetter_authNetrc(t *testing.T) {
 	mainPath := filepath.Join(dst, "main.tf")
 	if _, err := os.Stat(mainPath); err != nil {
 		t.Fatalf("err: %s", err)
+	}
+}
+
+func TestHttpGetter_GetByteRange(t *testing.T) {
+	var cases = []struct {
+		Input     string // example: "5555-66666"
+		Expected  []string
+		ExpectErr bool
+	}{
+		// both start and end bytes given
+		{"http://my/file.iso?ranged_request_bytes=555-666", []string{"555", "666"}, false},
+		// no end given
+		{"http://my/file.iso?ranged_request_bytes=555-", []string{"555", ""}, false},
+		// end greater than start
+		{"http://my/file.iso?ranged_request_bytes=666-555", nil, true},
+		// unparseable into ints
+		{"http://my/file.iso?ranged_request_bytes=66a6-5*55", nil, true},
+		// no ranged request made
+		{"http://my/file.iso", nil, false},
+	}
+
+	for _, testCase := range cases {
+		src := testCase.Input
+		u, err := urlhelper.Parse(src)
+		byteRange, err := getByteRange(u)
+
+		if byteRange == nil {
+			if testCase.Expected != nil {
+				t.Fatalf("unexpected output from getByteRange: expected %s; got %s",
+					testCase.Expected, byteRange)
+			}
+		} else if byteRange[0] != testCase.Expected[0] && byteRange[1] != testCase.Expected[1] {
+			t.Fatalf("unexpected output from getByteRange: expected %s; got %s",
+				testCase.Expected, byteRange)
+		}
+		if err == nil && testCase.ExpectErr {
+			t.Fatalf("Expected an error but didn't get one: %v", testCase)
+		}
+		if err != nil && !testCase.ExpectErr {
+			t.Fatalf("Got an error we were not expecting: %v, %s", testCase,
+				err)
+		}
 	}
 }
 
@@ -310,7 +388,7 @@ func testHttpHandlerExpectHeader(w http.ResponseWriter, r *http.Request) {
 }
 
 func testHttpHandlerFile(w http.ResponseWriter, r *http.Request) {
-	w.Write([]byte("Hello\n"))
+	http.ServeContent(w, r, "hello.txt", time.Now(), strings.NewReader("Hello\n"))
 }
 
 func testHttpHandlerHeader(w http.ResponseWriter, r *http.Request) {


SwampDragons:
Is it possible that https://github.com/hashicorp/go-getter/commit/4751e7c63c0ed7b55d8104884ea7c09fc7d677ab broke tests for windows?  I'm not sure how my change could be affecting this particular check.

SwampDragons:
@ryanuber @mitchellh I'm happy to fix these windows tests that are broken but I'm failing to understand which is failing -- the test or the underlying behavior.  What is _supposed_ to be happening here?  What are we testing? It looks like the test expects to have the pwd stripped from the url but at the same path the detector is full of checks to make sure it's returning an absolute path.

get_http.go
```diff
@@ -117,13 +118,54 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+	if err != nil {
+		if strings.Compare("No byte range provided", err.Error()) != 0 {
+			fmt.Printf("%s; going to download entire file without a range request",
+				err)
+		}
+	}
+
+	req, err := http.NewRequest("HEAD", u.String(), nil)
+
+	if len(byte_range) == 2 {
```
mwhooker: might make sense to have `GetByteRange` return an error if it didn't find 2 numbers?

get_http.go
```diff
@@ -117,13 +118,54 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
 		g.Client = httpClient
 	}
 
-	resp, err := g.Client.Get(u.String())
+	// check to see whether user has specified a range of bytes to download.
+	// if user has, but the range is invalid, fall back to downloading the
+	// whole file
+	isPartialDownload := false
```
mwhooker: this is a trivial comment, but it might make sense to move this closer down to where it's being used.

get_http.go
SwampDragons: already does; I guess I could just use this error instead of re-checking though.

mwhooker: ah, I see, this seems fine then. Could probably combine this code with the err checking above, but that's also trivial

get_http.go
SwampDragons: I guess I could move it down to ~ L134 if you think it'll make it read clearer...

get_http.go
```diff
@@ -220,6 +262,44 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
 	}
 }
 
+func GetByteRange(u *url.URL) ([]string, error) {
```
mitchellh: 1. I think this can be unexported `getByteRange` so we don't have to commit to public API stability of this, and you may want to consider attaching it to `(*HttpGetter)`.

2. I'd also document this method (even unexported). It'd help to note the precondition that `u` is non-nil and is also not modified. It'd also help to note the response permutations: `(nil, nil)`, `(nil, non-nil)`, `(non-nil, non-nil)`, `(non-nil, nil)`. Not all may be possible but usually my primary question with APIs in Go. In this case I think the pertinent ones are `(nil, nil) - no byte range specified, no error`, `(nil, non-nil) - byte range maybe specified, error`, `(non-nil, nil) - byte range with len == 2, idx0 < idx1`.

Note the quality of #2 is highly inconsistent across many Go codebases, but when I review I try to help champion some high quality comments. :)

get_http.go
```diff
@@ -220,6 +262,44 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
...
+	// format can be bytes=startByte- OR bytes=startByte-endByte
+	// example URL string: "http://my/file.iso?ranged_request_bytes=5555-66666"
+	// or  "http://my/file.iso?ranged_request_bytes=5555-"
+	// example bytes_range_retval: []int{5555, 66666} or []int{5555}
+	bytes_range_retval := make([]string, 0)
```
mitchellh: I'd remove this. Unless I'm mistaken, it isn't assigned anywhere and a `nil` response functionally is equivalent and more idiomatic for no result but also errors. This would also avoid an allocation (though that's really an unnecessary optimization and not really the point).

get_http.go
```diff
@@ -220,6 +262,44 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
...
+	q := u.Query()
+	byteRange := q.Get("ranged_request_bytes")
+	if byteRange == "" {
+		return bytes_range_retval, fmt.Errorf("No byte range provided")
```
mitchellh: I recommend returning `nil, nil`. This is idiomatic for "no result, but no error" a.k.a. "nothing to do." It appears this error is only string compared above anyways, so that'd turn that into a much more robust nil check, too!

get_http_test.go
```diff
@@ -207,6 +209,22 @@ func TestHttpGetter_authNetrc(t *testing.T) {
 	}
 }
 
+func TestHttpGetter_GetByteRange(t *testing.T) {
```
mitchellh: Happy to see tests 😄 I'd turn this into a table test, something of this structure:

```go
cases := []struct {
    Input string // example: "5555-66666"
    Expected []string
    Err bool
}
```

Then we can test all the edges and all the error branches in the func.

get_http.go
```diff
@@ -220,6 +262,44 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
...
+	errMsg := fmt.Errorf("Invalid byte range provided")
+	vals := strings.Split(byteRange, "-")
+
+	// validate byte range values given
+	if len(vals) != 2 {
+		return bytes_range_retval, errMsg
+	}
+	start, err := strconv.ParseInt(vals[0], 10, 64)
+	if err != nil {
+		return bytes_range_retval, errMsg
```
mitchellh: Returning `errMsg` alone hides what the error was, I'd recommend wrapping a new error `fmt.Errorf("Invalid byte range provided; %s", err)`.

get_http.go
```diff
@@ -220,6 +262,44 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
...
+	// example URL string: "http://my/file.iso?ranged_request_bytes=5555-66666"
+	// or  "http://my/file.iso?ranged_request_bytes=5555-"
+	// example bytes_range_retval: []int{5555, 66666} or []int{5555}
+	bytes_range_retval := make([]string, 0)
+	q := u.Query()
+	byteRange := q.Get("ranged_request_bytes")
+	if byteRange == "" {
+		return bytes_range_retval, fmt.Errorf("No byte range provided")
+	}
+	errMsg := fmt.Errorf("Invalid byte range provided")
```
mitchellh: Since this is constant and reused multiple times, I recommend moving it to a constant var at the bottom of the file in package scope.

```
var errInvalidByteRange = fmt.Errorf("Invalid byte range provided")
```

get_http.go
```diff
@@ -117,13 +118,54 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+	// check to see whether user has specified a range of bytes to download.
+	// if user has, but the range is invalid, fall back to downloading the
+	// whole file
+	byteRange, rangeErr := GetByteRange(u)
+	if rangeErr != nil {
+		if strings.Compare("No byte range provided", rangeErr.Error()) != 0 {
+			fmt.Printf("%s; going to download entire file without a range request",
+				rangeErr)
+		}
+	}
```
mitchellh: I'd remove the entire if statement above and invert the logging logic: only log when we _do_ specify a range request. That is the far more unlikely scenario (most HTTP requests do not have range requests) so it will result in a less noisy log I believe.

Also for logging, I think we should be using `log` here. I don't _think_ this package is updated to `go-hclog`, so its probably `log`. I'd just grep the package and see what the style is... I can't remember off the top of my head sorry.

get_http.go
```diff
@@ -117,13 +118,54 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+
+	req, err := http.NewRequest("HEAD", u.String(), nil)
```
mitchellh: Took me a second to realize so I'd comment here that you're allocating this up front because if it is a range request, we need to set `Range` headers on a shared request object.

get_http.go
```diff
@@ -117,13 +118,54 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+
+	isPartialDownload := false
+	if rangeErr == nil {
+		// We first make a HEAD request so we can check if the server supports
+		// range queries. If the server/URL doesn't support HEAD requests,
+		// we just fall back to GET.
+
+		if err != nil {
```
mitchellh: I'd move this entire `err` check up to when `NewRequest` is called. Under the current logic I think there is a edge case where: 

  * No byte range: `rangeErr != nil` under current logic
  * Request allocation fails `err != nil` 
  
And no error is reported since we will do a `g.Client.Do` on the GET below.

get_http.go
```diff
@@ -117,13 +118,54 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+		resp, err := g.Client.Do(req)
+		if err == nil && (resp.StatusCode >= 200 && resp.StatusCode < 300) {
+			// If the HEAD request succeeded, then attempt to set the range
+			// query if we can.
+			if resp.Header.Get("Accept-Ranges") == "bytes" {
+				req.Header.Set("Range", fmt.Sprintf("bytes=%s-%s",
+					byteRange[0], byteRange[1]))
+				isPartialDownload = true
+			}
+		}
```
mitchellh: What is the HEAD request failed? What if `err != nil` here? If we intend to ignore errors and download the full file, we should probably at least log that the HEAD request failed so range request will be ignored.

get_http.go
```diff
@@ -117,13 +118,54 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+		// We first make a HEAD request so we can check if the server supports
+		// range queries. If the server/URL doesn't support HEAD requests,
+		// we just fall back to GET.
+
+		if err != nil {
+			return err
+		}
+
+		resp, err := g.Client.Do(req)
+		if err == nil && (resp.StatusCode >= 200 && resp.StatusCode < 300) {
```
mitchellh: Parens unnecessary

get_http.go
```diff
@@ -117,13 +118,54 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+
+	resp, err := g.Client.Do(req)
 	if err != nil {
 		return err
 	}
 	defer resp.Body.Close()
 	if resp.StatusCode != 200 {
-		return fmt.Errorf("bad response code: %d", resp.StatusCode)
+		// if we did a ranged request we'll get a 206 instead.
+		if !((resp.StatusCode == 206) && (isPartialDownload == true)) {
```
mitchellh: Complex boolean logic (not your fault, just the fact of it!). Great job commenting it to help clarify, but may I suggest making a singular if with the following?

```go
if resp.StatusCode != 200 && !(isPartialDownload && resp.StatusCode == 206)`
```

I think it reads better literally: "if not a 200 response and its not a partial download with a 206". You can further demorgan's the not into the parens but as you did above I think its easier to read in english as the above...

I **think** thats equivalent logic but like I said, tricky logic so I could be wrong!

mitchellh:
Also yes, the windows failures are unrelated to this PR. So no worries there. We'll have to look at those separately sometime soon...

get_http.go
SwampDragons: This is unclear and I'll try to clean it up so it reads better but the current logic is that it only logs if the user provided a byte range but the code was unable to parse it, resulting in the range request being ignore and the entire file being downloaded.  I can log this as a warning instead though.

I'll replace this with the log package.

mitchellh: Ah sorry, it was late so unsurprising I made a mistake! In that case with the other suggestions in this PR I'd say we do a basic `if err != nil` and use `err == nil && result == nil` to specify no range request.

SwampDragons: Sounds good.  I'll noodle on this today.

get_http.go
SwampDragons: Document where?

mitchellh: Above, as a standard godoc-ish comment above the function.

The note I made was that as an unexported function godoc wouldn't show it of course but I still find that documenting methods is an important practice.

get_http.go
SwampDragons: oh yeah you're right.  That must have been an accident of refactoring because it makes no sense where I have it now.

SwampDragons:
I think I've made all the requested changes.

SwampDragons:
just rebased and resolved conflicts.

get_http.go
```diff
@@ -223,6 +266,63 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
...
+// (non-nil, non-nil) - should not occur
+// (nil, nil) - returned when no byte range was provided in the first place
+// (nil, non-nil) - error occured when parsing or validating the byte range
+// (non-nil, nil) - byte range was successfully parsed, and the range is
+// returned as an array of strings in the following format:
+// example non-nil return val "vals":
+// []string{5555, 66666}      // first url example above
+// []string{5555}             // second url example above
+func getByteRange(u *url.URL) ([]string, error) {
+	q := u.Query()
```
ryanuber: I'd probably just directly `u.Query().Get("ranged_request_bytes")` since you never use `q` again thereafter.

get_http.go
```diff
@@ -223,6 +266,63 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
...
+	vals := strings.Split(byteRange, "-")
+
+	// validate byte range values given
+	if len(vals) != 2 {
+		return nil, fmt.Errorf("%s; length of parsed byte range is not 2, %v",
+			invalidRangeMsg, vals)
+	}
+	start, err := strconv.ParseInt(vals[0], 10, 64)
+	if err != nil {
+		return nil, fmt.Errorf("%s; could not convert start byte string to an integer",
```
ryanuber: Can you include the original error in the message?

get_http.go
```diff
@@ -223,6 +266,63 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
...
+			invalidRangeMsg)
+	}
+	// If you leave endBytes blank, read to end of file.
+	if vals[1] != "" {
+		// otherwise make sure "finish" value is a number
+		finish, err := strconv.ParseInt(vals[1], 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("%s; could not convert finish byte string to an integer",
```
ryanuber: Same as above, could you embed the `err` to give the full reason of what went wrong? I like that you are setting context of the error in here though :+1:

get_http_test.go
```diff
@@ -159,6 +163,38 @@ func TestHttpGetter_file(t *testing.T) {
 	assertContents(t, dst, "Hello\n")
 }
 
+var rangeReqs = []struct {
```
ryanuber: I would put this directly in your test case to keep it scoped to just that test.

get_http_test.go
```diff
@@ -215,6 +251,48 @@ func TestHttpGetter_authNetrc(t *testing.T) {
 	}
 }
 
+var cases = []struct {
```
ryanuber: Same as above - this should probably go inside of the test it is used for, since we aren't using it anywhere else other than that test.

get_http.go
```diff
@@ -118,12 +118,55 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+				isPartialDownload = true
+			}
+		} else {
+			log.Printf("HEAD request for Range failed; falling back to full file download: %s",
+				err)
+		}
+	}
+
+	// Set the request to GET now, and redo the query to download
+	req.Method = "GET"
```
ryanuber: I'm not actually sure about the API promise if you reuse an `*http.Request` struct. I would probably just create a new request here to be safe. As an example of why I would worry about this, the request's `Body` gets automatically closed by the transport. To me that implies that an `*http.Request` is not meant to be reused for multiple requests, even though it might work in this case.

get_http.go
```diff
@@ -118,12 +118,55 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+		// range queries. If the server/URL doesn't support HEAD requests,
+		// we just fall back to GET.
+
+		resp, err := g.Client.Do(req)
+		if err == nil && resp.StatusCode >= 200 && resp.StatusCode < 300 {
+			// If the HEAD request succeeded, then attempt to set the range
+			// query if we can.
+			if resp.Header.Get("Accept-Ranges") == "bytes" {
+				req.Header.Set("Range", fmt.Sprintf("bytes=%s-%s",
+					byteRange[0], byteRange[1]))
```
ryanuber: Won't `byteRange[1]` be out of range for cases where `byteRange` is just a single element? In your comment below you mentioned this as a possible return value from `getByteRange()`: `[]string{5555}`.

ryanuber: Ah, I see, I think the code comment is just outdated. Looks like you directly return the result of `strings.Split()` which would have an empty string as the second element.

get_http.go
```diff
@@ -223,6 +266,63 @@ func (g *HttpGetter) parseMeta(r io.Reader) (string, error) {
...
+// note that the format of the raw url string can be either
+// bytes=startByte- OR bytes=startByte-endByte
+//
+// return values:
+// (non-nil, non-nil) - should not occur
+// (nil, nil) - returned when no byte range was provided in the first place
+// (nil, non-nil) - error occured when parsing or validating the byte range
+// (non-nil, nil) - byte range was successfully parsed, and the range is
+// returned as an array of strings in the following format:
+// example non-nil return val "vals":
```
ryanuber: 🤔 I'm not sure what line 286 means?

README.md
```diff
@@ -256,6 +256,10 @@ None
   * `rev` - The Mercurial revision to checkout.
 
 ### HTTP (`http`)
+  * `ranged_request_bytes` - If the server supports range requests, this will 
+  set the `Range` header when making the HTTP request, allowing you to specify 
+  a specific range of bytes to download.
+  Example usage: `http://www.path/2/my/file.iso?ranged_request_bytes=5555-66666`
```
ryanuber: Can we also include an example/explanation of omitting the second parameter in the range?

get_http.go
SwampDragons: lol the name of the value we're returning is "vals" but you're right that reads like absolute gibberish.

get_http.go
SwampDragons: Will fix comment.

get_http_test.go
SwampDragons: Can do. Makes sense.

SwampDragons:
@ryanuber nudge. I know we discussed having someone else review this but I only just got back to looking at it and whaddya know you're back from your leave.

get_http.go
```diff
@@ -118,12 +120,60 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+
+	// First request here (preReq), is going to be used to determine whether we
+	// can make a ranged request. Second one, (req) is used for the actual
+	// download
+	preReq, err := http.NewRequest("HEAD", u.String(), nil)
+	if err != nil {
+		return err
+	}
+
+	req, err := http.NewRequest("HEAD", u.String(), nil)
```
ryanuber: I think you can just set `"GET"` all the way up here now that you have two separate HTTP calls, and remove the line below that sets this!

get_http.go
```diff
@@ -118,12 +120,60 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+		// log that we are downloading whole file even though user
+		// requested a byte range.
+		log.Printf("%s; going to disregard range request and download entire file",
+			rangeErr)
+	}
+
+	// First request here (preReq), is going to be used to determine whether we
+	// can make a ranged request. Second one, (req) is used for the actual
+	// download
+	preReq, err := http.NewRequest("HEAD", u.String(), nil)
```
ryanuber: This is totally fine and works as is, but you could alternatively use the shorthand for this: `resp, err := g.Client.Head(u.String())` to tighten this up a bit.

get_http.go
```diff
@@ -118,12 +120,60 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
...
+	if err != nil {
+		return err
+	}
+
+	isPartialDownload := false
+	if rangeErr == nil && byteRange != nil {
+		// We first make a HEAD request so we can check if the server supports
+		// range queries. If the server/URL doesn't support HEAD requests,
+		// we just fall back to GET.
+		resp, err := g.Client.Do(preReq)
```
ryanuber: If this `err` comes back nil, we should take care to close the response body to avoid leaking a file descriptor, as we do below on line 174 for the actual GET request. It looks like net/http doesn't do anything special for HEAD specifically, so we are still required to close the body ourselves. You can probably just immediately close it and move on in this case.

SwampDragons:
Hmm, just revisited this PR after letting it fester way too long and it looks like @vancluever's recent change allowing us to set headers actually renders it unnecessary. I'm gonna close this and see about using that for ranged requests instead, and do any error handling I need on the Packer side.

PR #96, opened 2018-01-24
feature: Support for maven type

missedone:


missedone:
@schmichael, I added test case for central maven repo and jcenter.

get_mvn.go
```diff
@@ -0,0 +1,161 @@
...
+		ver = snapshotVer
+	}
+
+	filename := artifactId + "-" + ver
+	if classifier != "" {
+		filename += "-" + classifier
+	}
+	filename += "." + artType
+	artifactUrl.Path += "/" + filename
+	dstFile := dst
```
schmichael: Same here. I don't see a reason for the secondary `dstFile` variable.

schmichael: Also is there a way to use a small helper func to format the filename since there's similar logic here and in Get?

get_mvn.go
```diff
@@ -0,0 +1,161 @@
...
+	// if it's not auto decompress archive mode, use the real file name
+	if !strings.HasSuffix(dst, "/archive") {
+		dstFile = filepath.Join(filepath.Dir(dst), filename)
+	}
+
+	log.Printf("Downloading %s to %s", artifactUrl, dstFile)
```
schmichael: Remove. We don't log from this library.

get_mvn.go
```diff
@@ -0,0 +1,161 @@
...
+	return g.HttpGet.GetFile(dstFile, artifactUrl)
+}
+
+func (g *MvnGetter) ParseLastestSnapshotVersion(artifactUrl *url.URL) (string, error) {
```
schmichael: I don't think there's a reason to export this, so please lowercase the first letter.

get_mvn.go
```diff
@@ -0,0 +1,161 @@
...
+	mvnMetaUrl, err := url.Parse(artifactUrl.String())
+	if err != nil {
+		return "", err
+	}
+	mvnMetaUrl.Path += "/maven-metadata.xml"
```
schmichael: Do you need to do the same `mvnMetaUrl.RawPath = ""` here? This pattern should probably have the same behavior in both locations.

get_mvn.go
```diff
@@ -0,0 +1,161 @@
...
+
+	mvnMetaFile, err := ioutil.TempFile(os.TempDir(), "maven-metadata")
```
schmichael: You can just use an empty string instead of TempDir and TempFile will do the right thing: `ioutil.TempFile("", ...)`

missedone:
@schmichael , I updated the PR incorporate with your review comments, thanks.

be noted that, this time I introduced a new api Getter.Getfilename to allow Getters override the default filename, it's useful especially for Maven support that the default filename comes with the specified artifactId, version and type.
the original logic use base url.Path as the default filename, while it's not fit for maven or other case that that not reflect the name in url.Path, also specify query parameter 'filename' for each url is sort of redundant:
```golang
				filename = filepath.Base(u.Path)

				// Determine if we have a custom file name
				if v := q.Get("filename"); v != "" {
					// Delete the query parameter if we have it.
					q.Del("filename")
					u.RawQuery = q.Encode()

					filename = v
				}
```

client.go
```diff
@@ -232,15 +232,22 @@ func (c *Client) Get() error {
 		// Destination is the base name of the URL path in "any" mode when
 		// a file source is detected.
 		if mode == ClientModeFile {
-			filename := filepath.Base(u.Path)
+			filename, err := g.GetFilename(u)
```
missedone: @schmichael , here use the new api Getter.Getfilename to allow Getters override the default filename.

by default `GetFilename` returns "" so that the other Getters fall back to original logic of computing filename

get_mvn.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	filename += "." + artType
+	artifactUrl.Path = path.Join(artifactUrl.Path, filename)
+
+	return g.HttpGet.GetFile(dst, artifactUrl)
+}
+
+// get the latest snapshot version by parsig the maven-metadata.xml from remote maven repo.
+//   - artifactVerUrl the url to the artifact version, Ex., 'https://repo1.maven.org/maven2/org/testng/testng/6.13.1/'
+func (g *MvnGetter) ParseLastestSnapshotVersion(artifactVerUrl *url.URL) (string, error) {
```
missedone: @schmichael , this method `ParseLastestSnapshotVersion ` is used by the test code, so that's why it's exposed.

get_mvn.go
```diff
@@ -0,0 +1,161 @@
...
+	artType := q.Get("type")
+	if artType == "" {
+		artType = "jar"
+	}
+
+	filename := artifactId + "-" + version
+	if classifier != "" {
+		filename += "-" + classifier
+	}
+	filename += "." + artType
```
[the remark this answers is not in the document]
missedone: `filename` is the name of the file, not relate to file path, so string concatenation is OK here.

missedone:
@schmichael, could you review the updates and hopefully merge it? Thanks.

get_mvn.go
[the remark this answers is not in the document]
schmichael: It's constructed out of arbitrary query parameters and appended to a path, so I think we should clean it.

For example if the URL contains something like `?artifactId=../foo` it would get joined to `dstDir` and end up escaping `dstDir`.

This assumes the Nomad situation where `dstDir` is controlled by an application (Nomad) but the URL is accepted as user input. Since the URL is user input we need to make sure it doesn't escape `dstDir`.

get_mvn.go
schmichael: Since the test is in the same package this does not need to be exported.

client.go
schmichael: Argh, sorry if I misled you, but I don't think adding a new method to the interface (and therefore every single implementation) is worth it if it's only implemented by the Maven getter.

get_mvn.go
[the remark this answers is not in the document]
missedone: ok, i see, thanks for the detail explanation

client.go
missedone: I think add this new method `GetFilename ` leaves flexibility for future in case new Getter need to control the default filename rather than passing query parameter "filename".

also, my original implementation of Getter.GetFile(dst, url) did not comply with the api as it does not save file to `dst`, instead, it's "dir(dst)/artifactId-version.type".
so two solution to fix this:
1) pass the correct filename in dst by passing the query parameter 'filename' in each url. (this is what i'm trying to avoid, especially when the default filename can be easily inferred by the MvnGetter)
2) do something in the interface to allow Getter control the default filename, the disadvantage is need to add single implementation to each Getter since golang not like java that support abstract class)

missedone: @schmichael, what do you think about this?

mpilone:
What would it take to get this feature/PR moving again? We have a need to pull SNAPSHOTs from a Maven repo in our test environments and this feature would allow us to avoid writing a proxy HTTP server in front of out Nexus repository.

It looks like there are now conflicts/build failures because the PR has been sitting for a while. But the bigger concern is it looks like the contributor license hasn't been signed so I'm assuming this feature can't be merged until that is done?

CreatorHead:
Thanks for the contribution and for taking the time to propose Maven support.

This PR has been open since 2018 and no longer applies cleanly to the current codebase. It also introduces a new getter/backend, which adds long-term maintenance and security surface area.

There’s an additional concern noted in the review discussion: the output filename is constructed from user-controlled query parameters, which can potentially allow path traversal outside the destination directory (for example via artifactId=../foo). That risk would need to be addressed very carefully with strict validation and tests.

To keep the repository healthy and manageable, we’re going to close this PR.
If Maven support is still needed, please open a new PR based on current main, with CI passing, clear input validation, and tests demonstrating that user input cannot escape the destination directory.

Thanks again for the effort and for understanding.

PR #97, opened 2018-01-24
add simple sftp support

missedone:
for #7

### DIFF
--- a/get_sftp_test.go
+++ b/get_sftp_test.go
@@ -1,0 +1,43 @@
+package getter
+
+import (
+	"io/ioutil"
+	"net/url"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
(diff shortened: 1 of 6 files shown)


get_sftp_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestSftpGetter_impl(t *testing.T) {
+	var _ Getter = new(SftpGetter)
+}
+
+func TestSftpGetter_file(t *testing.T) {
+	g := new(SftpGetter)
+	dst := filepath.Join(tempDir(t), "readme.txt")
+
+	u, err := url.Parse("sftp://demo:password@test.rebex.net:22/readme.txt")
```
missedone: this is free public sftp server for testing:
http://www.sftp.net/public-online-sftp-servers

I'm not able to programatically setup a mock sftp server for unit testing in short, so just heading to this quick but dirty way.

PR #100, opened 2018-02-14
Clean up temporary file generated by get_http_test

godef:


mitchellh:
Thanks, looks good to me.

PR #101, opened 2018-02-15
Check File.Close error for get_http.go 

godef:
File.Close() could fail. According to close() system call, the following error could happen:
       EBADF  fd isn't a valid open file descriptor.
       EINTR  The close() call was interrupted by a signal; see signal(7).
       EIO    An I/O error occurred.

### DIFF
--- a/get_http.go
+++ b/get_http.go
@@ -135,9 +135,14 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
 	if err != nil {
 		return err
 	}
-	defer f.Close()
 
-	_, err = io.Copy(f, resp.Body)
+	n, err := io.Copy(f, resp.Body)
+	if err == nil && n < resp.ContentLength {
+		err = io.ErrShortWrite
+	}
+	if err1 := f.Close(); err == nil {
+		err = err1
+	}
 	return err
 }
 


mitchellh:
Thanks! The logic in this unfortunately isn't quite right, since if it there is an error during `io.Copy` the file is never closed. If you fix that logic up then I'd be happy to merge this.

godef:
Yes, that's good catch. fixed. Could you take another look? Thanks so much.

mwhooker:
Looking at [`ioutil.WriteFile`](https://github.com/golang/go/blob/1d07ed15798b1228a2777c8ad80353b17a9ad8bd/src/io/ioutil/ioutil.go#L79-L91) as an example,

what do you think about something like

```diff
diff --git a/get_http.go b/get_http.go
index 9acc72c..720338a 100644
--- a/get_http.go
+++ b/get_http.go
@@ -135,9 +135,16 @@ func (g *HttpGetter) GetFile(dst string, u *url.URL) error {
        if err != nil {
                return err
        }
-       defer f.Close()

-       _, err = io.Copy(f, resp.Body)
+       n, err := io.Copy(f, resp.Body)
+
+       if err == nil && n < resp.ContentLength {
+               err = io.ErrShortWrite
+       }
+
+       if err1 := f.Close(); err == nil {
+               err = err1
+       }
        return err
 }
```

godef:
I think that's a good idea. Yes, I also looked at WriteFile before, but thought io.Copy will return error if the write length is wrong. I think it's safer to check that, and it also checks the resp.ContentLength vs the actually data, which is good. Thanks :-)

godef:
ping :-)

godef:
ping. Who can help to merge this to master? :-)

schmichael:
Sorry for the delay and thanks for the patch @godef !

PR #103, opened 2018-04-10
Apply mode flags on directories

angrycub:
Repair the directory permissions from the archive while restoring the mtime.  This is necessary because we inject artificial permissions when creating a directory.


decompress_tar.go
```diff
@@ -105,6 +105,10 @@ func untar(input io.Reader, dst, src string, dir bool) error {
 	// We therefore wait until we've extracted everything and then set the mtime and atime attributes
```
schmichael: Can we update the comments to be something like:

```go
// Perform a final pass over extracted directories to update metadata
for ... {
  ...
  // Chmod the directory...
  ...

  // Set the mtime/atime attributes since they would have been changed during extraction
  ...
}
```

PR #106, opened 2018-04-25
Preserve accesstime and modifytime only when valid

preetapan:
This PR fixes a change in behavior introduced in https://github.com/hashicorp/go-getter/commit/0a8cdac4da1802174ae5302600ff948b6a855349 that preserves access time and modified time. This causes problems with usages of go getter in environments where the files were created with 
 no access time, reported by a user in https://github.com/hashicorp/nomad/issues/4194. 

Instead, we set the access/modify time to current time if its <=0, so that its less distruptive to client apps that rely on go-getterh

cc @dadgar @chelseakomlo

### DIFF
--- a/decompress_tar.go
+++ b/decompress_tar.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"time"
 )
 
 // untar is a shared helper for untarring an archive. The reader should provide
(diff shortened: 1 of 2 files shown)


decompress_tar.go
```diff
@@ -90,8 +91,16 @@ func untar(input io.Reader, dst, src string, dir bool) error {
 			return err
 		}
 
-		// Set the access and modification time
-		if err := os.Chtimes(path, hdr.AccessTime, hdr.ModTime); err != nil {
+		// Set the access and modification time if valid, otherwise default to current time
+		aTime := time.Now()
```
dadgar: Capture now outside the for loop (https://github.com/hashicorp/go-getter/pull/106/files#diff-0d8d3e8dc13e64092481d7dd63191a0fR17) and use it here and at https://github.com/hashicorp/go-getter/pull/106/files#diff-0d8d3e8dc13e64092481d7dd63191a0fR112.

decompress_tar.go
```diff
@@ -90,8 +91,16 @@ func untar(input io.Reader, dst, src string, dir bool) error {
...
+		mTime := time.Now()
```
mitchellh: These two lines could result in two different values, can we do `mTime := aTime` instead. `time.Time` is just a value, so it'll copy.

PR #111, opened 2018-07-09
Fix git getter tests + a minor bug fix

ryanuber:
Fixes the SSH key auto-disabling behavior we have been seeing (hopefully), and fixes a minor bug in the git getter where the output of `git version` contains some extra build metadata, as recent versions seem to have.


ryanuber:
@jkohrman I thought about that, but thought that it would be more likely to be parseable if it were base64 by a scanner, since base64 will give you predictable text for the header of an SSH key. I thought splitting it right in the middle with some language-specific concatenation would make it harder for a scanner to piece back together. I suppose we could do both, but not sure there is any gain. What do you think?

jkohrman:
No tremendous gain either way.  I'm leery of string concatenation for inputs like this, but this is a fair breakfix.  Let's consider non-plaintext if it happens again.

PR #113, opened 2018-08-09
Use a new AWS access key

jfreda:
The current AWS access key was flagged and we need to rotate it. This access key is locked down with the same IAM permissions. I looked into a couple of ways that we might be able to avoid this issue in the future, but they aren't trivial so this is at least the temporary fix.


ryanuber:
Dang, yeah. It's a little surprising that even with concatenation this gets flagged. :+1: to get things working again.

PR #115, opened 2018-09-13
release binaries to github

moshloop:
This PR will automatically publish binaries to Github for every tag see https://github.com/moshloop/go-getter/releases

This makes installing and using go-getter as a cli tool much easier

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: true
 
 addons:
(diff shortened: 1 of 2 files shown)


sj26:
This is the first thing I looked for when discovering this project — please provide binaries I can drop into my container images 🙇‍♂️

philomory:
Is there any chance of movement on this?

PR #125, opened 2018-11-14
don't get file when checksum of destination is valid

azr:
Howdy,

It's all in the title.

This is to avoid downloading the file if it is already there and the checksum is valid.

This will help a lot in packer as we download big isos in a cache directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,9 @@ is shown below:
 The checksum query parameter is never sent to the backend protocol
 implementation. It is used at a higher level by go-getter itself.
 
+If the destination file exists and the checksums match: download
+will be skipped.
+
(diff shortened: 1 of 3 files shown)


azr:
I was just thinking the same, adding them ! 🙂

azr:
Here we go 🙂

azr:
Hey @mitchellh, is it better now ? 🙂

vancluever:
Hey @azr, I just updated this so that the test failure case worked (it was causing a nil panic when you remove the checksum as an ad hoc failure test).

Otherwise, looks good! I'll merge it shortly.

azr:
Awesome, thanks 🙂

PR #129, opened 2018-11-17
Allow SCP-style URLs for Git

mitchellh:
This fixes both #38 and #124, and should also address the 4 TF issues referenced there as well. First, thanks @paultyng for the failing cases as well as a stab at a solution. You can see I took from that! I also took the failing cases from #38 (now passing). 

I dug into the linked Terraform issues and there were actually two issues at play, both fixed in this PR:

**1.) Detector for SCP-like Git URLs:** 

The heuristic we use here is the `git@` username. We can extend this to support more usernames in the future (the code is ready for it) but all the issues seem to use this as the user and this limits the blast radius of this change significantly. 

This detects things in the form of `git@foo.com:path` and converts it to `git::ssh://git@foo.com/path` and Git will clone this properly. 

Note that if a user other than "git" is required, the go-getter user can still use full forced URLs like `git::ssh://admin@host.com/path`. 

**2.) Fixing up the SCP-like URLs directly in the Git getter:** 

If a full, valid URL is given to go-getter, then the detectors are not run. So if a user gave the URL `git::ssh://git@foo.com:path` (note the colon), then the detector won't run to fix that up at all. Therefore, I also added a little fixup directly in the Git getter that looks for 1.) an SSH URL and 2.) with ":" in the host (how Go's `net/url` parses it) and fixes it up.

This should fix the issues where the direct URL is used.

**Note to @apparentlymart:** Re: your comment, I verified that forced protocols will always force. This is done here: https://github.com/hashicorp/go-getter/blob/9fd5bfc48c2d3bc3a8251bc9c8dba824d1b02dab/detect.go#L93 So your comment should be okay, everything works as you would expect.

## Backwards Compatibility

This should not break any backwards compatibility.

These SCP-like URLs were being detected as "file" URLs below and its highly unlikely that was ever the intended behavior. Users who may have intended this can force it with `file::`, though I strongly believe the number of users impacted by that will be zero.

## Future Stuff

The groundwork is now laid (`detect_ssh.go`) to support an SCP-getter directly as well. So we can consider doing that one day by expanding this detector directly into SSH URLs that use an SCP lib or something.

### DIFF
--- a/detect_git_test.go
+++ b/detect_git_test.go
@@ -1,0 +1,38 @@
+package getter
+
+import (
+	"testing"
+)
+
+func TestGitDetector(t *testing.T) {
+	cases := []struct {
+		Input  string
+		Output string
+	}{
+		{"git@github.com:hashicorp/foo.git", "git::ssh://git@github.com/hashicorp/foo.git"},
+		{
+			"git@github.com:hashicorp/foo.git//bar",
+			"git::ssh://git@github.com/hashicorp/foo.git//bar",
+		},
+		{
+			"git@github.com:hashicorp/foo.git?foo=bar",
+			"git::ssh://git@github.com/hashicorp/foo.git?foo=bar",
+		},
+	}
+
+	pwd := "/pwd"
+	f := new(GitDetector)
+	for i, tc := range cases {
+		output, ok, err := f.Detect(tc.Input, pwd)
+		if err != nil {
+			t.Fatalf("err: %s", err)
+		}
+		if !ok {
+			t.Fatal("not ok")
+		}
+
+		if output != tc.Output {
+			t.Fatalf("%d: bad: %#v", i, output)
+		}
+	}
+}
--- a/detect_git_test.go
+++ b/detect_git_test.go
@@ -11,12 +11,32 @@ func TestGitDetector(t *testing.T) {
 	}{
 		{"git@github.com:hashicorp/foo.git", "git::ssh://git@github.com/hashicorp/foo.git"},
 		{
+			"git@github.com:org/project.git?ref=test-branch",
+			"git::ssh://git@github.com/org/project.git?ref=test-branch",
+		},
+		{
 			"git@github.com:hashicorp/foo.git//bar",
 			"git::ssh://git@github.com/hashicorp/foo.git//bar",
 		},
 		{
 			"git@github.com:hashicorp/foo.git?foo=bar",
 			"git::ssh://git@github.com/hashicorp/foo.git?foo=bar",
+		},
+		{
+			"git@github.xyz.com:org/project.git",
+			"git::ssh://git@github.xyz.com/org/project.git",
+		},
+		{
+			"git@github.xyz.com:org/project.git?ref=test-branch",
+			"git::ssh://git@github.xyz.com/org/project.git?ref=test-branch",
+		},
+		{
+			"git@github.xyz.com:org/project.git//module/a",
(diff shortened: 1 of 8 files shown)


detect_github_test.go
```diff
@@ -24,17 +24,6 @@ func TestGitHubDetector(t *testing.T) {
...
-		// SSH
-		{"git@github.com:hashicorp/foo.git", "git::ssh://git@github.com/hashicorp/foo.git"},
-		{
-			"git@github.com:hashicorp/foo.git//bar",
-			"git::ssh://git@github.com/hashicorp/foo.git//bar",
-		},
-		{
-			"git@github.com:hashicorp/foo.git?foo=bar",
-			"git::ssh://git@github.com/hashicorp/foo.git?foo=bar",
-		},
```
mitchellh: These tests were moved to `detect_git_test.go`, not removed.

detect_git_test.go
```diff
@@ -0,0 +1,58 @@
...
+			"git@github.xyz.com:org/project.git//module/a",
+			"git::ssh://git@github.xyz.com/org/project.git//module/a",
+		},
+		{
+			"git@github.xyz.com:org/project.git//module/a?ref=test-branch",
+			"git::ssh://git@github.xyz.com/org/project.git//module/a?ref=test-branch",
+		},
+	}
+
+	pwd := "/pwd"
```
paultyng: Not a big deal but to avoid future regression should you maybe also test with empty string?

get_git.go
```diff
@@ -78,6 +78,26 @@ func (g *GitGetter) Get(dst string, u *url.URL) error {
 		}
 	}
 
+	// For SSH-style URLs, if they use the SCP syntax of host:path, then
+	// the URL will be mangled. We detect that here and correct the path.
+	// Example: host:path/bar will turn into host/path/bar
+	if u.Scheme == "ssh" {
+		if idx := strings.Index(u.Host, ":"); idx > -1 {
```
paultyng: How does this interact with a non standard port for ssh? I think you would only want to do this if it is definitely scp style. May want to add some test cases for that.

detect_git_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"testing"
+)
+
+func TestGitDetector(t *testing.T) {
+	cases := []struct {
+		Input  string
+		Output string
+	}{
```
paultyng: Some systems like AWS code commit use non git user names: `git clone ssh://Your-SSH-Key-ID@git-codecommit.us-east-2.amazonaws.com/v1/repos/MyDemoRepo my-demo-repo`, not sure if its worth handling that in SCP style

get_git.go
mitchellh: It appears that if you want to specify the port they're mutually exclusive: https://stackoverflow.com/questions/5767850/git-on-custom-ssh-port

At the point this runs, the URL has already been parsed, so the port is on `u.Port`. I believe the only case that `:` would be present in `u.Host` is the SCP-like URL case.

detect_git_test.go
mitchellh: Yeah, for now you have just force it with `git::ssh://`, the autodetection won't work for custom usernames. We can relax this later but I wanted to limit blast radius to start, and there is a decent workaround as noted 😄

detect_git_test.go
mitchellh: In reality a PWD will almost always be given since go-getter always populates it, I think its safer from a test perspective to use a bogus one like this. If the detector ever uses the pwd, the tests should fail in either case.

PR #130, opened 2018-11-17
Add go.mod

mitchellh:
This adds a `go.mod` file.

If we find we need to lower the minimum version for downstream projects then we can easily do that. This gets us a list of deps that works (`go test` green) as is.


mitchellh:
Yeah I felt the same way (the AWS dep specifically). But I think we can safely lower the minimum version when that time comes if we have to!

PR #131, opened 2018-11-19
Client options + http progress + http byte range requets

azr:
Howdy,

This PR introduces a non breaking change that allows to configure a client by changing Get/GetAny/GetFile api:
```go
func Get(dst, src string, opts ...ClientOption) error {
```

This pr also introduces **http download continue** to save some bandwidth.

This pr also allows to request for a progress bar display. 
( which only works for http downloads for now )

In a future PR I would like to add context support ~& race protections~.


azr:
Ah, need to fix some tests first.

client_option_progress_cheggaaa.go
```diff
@@ -0,0 +1,67 @@
+package getter
+
+import (
+	"io"
+	"path/filepath"
+
+	pb "gopkg.in/cheggaaa/pb.v1"
```
vancluever: I think this is an issue for me in that that if we were to roll this progress bar in with go-getter, that would mean that _all_ systems that depend on go-getter would pull this in, even if they wanted to implement their own progress bar. I would prefer that this be removed, we keep the interface, and progress bars be explicitly implemented within each consuming implementation.

cmd/go-getter/main.go
```diff
@@ -10,6 +10,7 @@ import (
 
 func main() {
 	modeRaw := flag.String("mode", "any", "get mode (any, file, dir)")
+	progress := flag.Bool("progress", false, "display terminal progress")
```
vancluever: I see why `gopkg.in/cheggaaa/pb.v1` is being included now.

Can we move this down to this package (or even a package below `cmd/go-getter`) so that it's only used when the CLI needs it?

azr:
All right, thanks for reviewing !!
I moved the external dependency and added some tests for http progress. 🙂

azr:
Should I then remove the reference to `github.com/cheggaaa/pb` in `go.mod`

vancluever:
Hey @azr, getting close:

* Can you move the progress bar to the `cmd/go-getter` package? I don't think we need to have a separate package for it - it's a great implementation example but it still is really an example and hence does not need its own package that kind of indicates that it isn't. 
* Can you explain why the mutex was necessary? I think I know why just want to confirm.
* Finally, you don't need to modify `go.mod` to remove the dep, I've confirmed that it does not add it to any dependent packages.

Thanks!

azr:
Hey, @vancluever 🙂 
* about the progress bar implementation: I didn't think of it as an example, I put it there because I also use it packer side. It thought it was handy since the same implementation could be used in `cmd/go-getter` and in `packer`, I am totally fine copy pasting it but I was thinking that since we already import the go-getter repo from packer; it's not too bad. Do you confirm I should remove it ?
* The mutex is necessary because two downloads can start/happen/finish at the same time in a `packer build` for example and then this could race. Not sure if this can happen for the go-getter.
* Great ! 🙂 

Cheers !

vancluever:
@azr:

> about the progress bar implementation: I didn't think of it as an example, I put it there because I also use it packer side. It thought it was handy since the same implementation could be used in cmd/go-getter and in packer, I am totally fine copy pasting it but I was thinking that since we already import the go-getter repo from packer; it's not too bad. Do you confirm that I should remove it ?

Yeah, I think this dependency should be in packer specifically and not go-getter. Copy/paste is fine if you want, it's just the implementation should be purely local. Don't remove it from here, move it to the command package so it can be used there, but also include it in packer. This signals to each consumer that if they want a progress bar they need to use a local implementation, keeping the dependencies where they should be.

> The mutex is necessary because two downloads can start/happen/finish approximatively at the same time in a packer build for example and then this could race. Not sure if this can happen for the go-getter cmd.

I'm a bit confused here - so does Packer support parallel downloads with the same getter or are they serial, and you want to ensure a serial, synchronous behavior with some of the goroutines that are passed back? Just trying to understand things here. With the implementation being local, it probably doesn't matter so much, but I just want to understand what the expectation here is, especially for go-getter which I don't think is designed to be concurrent.

cmd/go-getter/progress_tracking.go
```diff
@@ -10,8 +10,7 @@ import (
 )
 
 // DefaultProgressBar is the default instance of a cheggaaa
-// progress bar. It is recommended to use DefaultProgressBar
-// in all places.
+// progress bar.
 var DefaultProgressBar getter.ProgressTracker = &ProgressBar{}
```
vancluever: Almost there. Just need to make this unexported now and we're good!

azr: Okay, done 🙂

sagikazarmark:
This is awesome! Can't wait to use it!

vancluever:
Looks good now yeah! Merging :+1:

azr:
Awesome !! 🎉  
thanks for your time :D

PR #134, opened 2018-11-21
Remove hardcoded master branch reference

### LINKED ISSUE #104: GitGetter update assumes a master branch is available
When calling [update](https://github.com/hashicorp/go-getter/blob/90bb99a48d86cf1d327cee9968f7428f90ba13c1/get_git.go#L147-L169) in a detached head repo (e.g. a Terraform module), `git checkout master` is always executed. This causes a `pathspec 'master' did not match any file(s) known to git` error to be raised during `terraform get` if a master branch is not available.

The `// Not a branch, switch to master...` logic should be updated to check for the default branch instead of assuming master is available. Something akin to `git branch -r --points-at refs/remotes/origin/HEAD | grep '\->' | cut -d' ' -f5 | cut -d/ -f2` ([sourced from SO](https://stackoverflow.com/a/42570510/2547802)) seems like a reliable option.

([This option](https://stackoverflow.com/a/49384283/2547802) seemed like it would be much cleaner, but unfortunately doesn't seem to work in detached HEAD)

troyready:
Fixes #104

### DIFF
--- a/get_git.go
+++ b/get_git.go
@@ -1,6 +1,7 @@
 package getter
 
 import (
+	"bytes"
 	"context"
 	"encoding/base64"
 	"fmt"
@@ -9,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


troyready:
Does seem safer -- updated the PR to fallback to `master`

troyready:
Rebased to fix conflicts. CI errors look unrelated to me.

troyready:
@radeksimko would you mind reviewing this? It's listed with red `X`s for CI but I think those were unrelated issues.

I could push a change to trigger CI again if that'd help?

troyready:
@azr would you mind reviewing this? I believe the CI failures are because of #210 

(sorry to bug you randomly, just trying to move this along but it's been stalled out for some time with the appearance of failing CI)

PR #136, opened 2018-12-03
go-get checksum file & guess checksum

azr:
Howdy,

This pr allows to checksum from a file; this should be the last needed feature for packer to be able to use go-getter 🙂 

`file_url?checksum=file:checksum_url`

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,15 +156,30 @@ a checksum for you. Note that checksumming only works for downloading files,
 not directories, but checksumming will work for any protocol.
 
 To checksum a file, append a `checksum` query parameter to the URL.
-The paramter value should be in the format of `type:value`, where
-type is "md5", "sha1", "sha256", or "sha512". The "value" should be
-the actual checksum value. go-getter will parse out this query parameter
-automatically and use it to verify the checksum. An example URL
-is shown below:
+The parameter value should be in the format of `type:value`, where
+type is "md5", "sha1", "sha256", "sha512" or "file". The "value" should be
+the actual checksum value or download URL for "file".
+go-getter will parse out this query parameter automatically
+and use it to verify the checksum. 
+Examples:
 
 ```
 ./foo.txt?checksum=md5:b7d96c89d09d9e204f5fedc4d5d55b21
 ```
+
+```
+./foo.txt?checksum=file:./foo.txt.sha256sum
+```
+ 
+When checksumming from a file, go-getter will go get checksum_url
+in a temporary directory and parse the content of the file.
+Content of files are expected to be BSD style or GNU style.
+For GNU-style checksum files; it is very common that the hashing algorithm identifier
+is in the filename; so the name of every supported hashing algorithm is compared
+against checksum_url for a match/guess.
+In case a different hashing algorithm is in the filename of checksum_url
+it is recommended to explicitly set hashing algorithm instead; the
+outcome would be undetermined otherwise.
 
 The checksum query parameter is never sent to the backend protocol
 implementation. It is used at a higher level by go-getter itself.
--- a/checksum.go
+++ b/checksum.go
@@ -66,7 +66,7 @@ func checksumHashAndValue(u *url.URL) (checksumHash hash.Hash, checksumValue []b
 	}
 
 	return nil, nil, fmt.Errorf(
-		"Could not find a supported checksum in %s: %s", file, checksumType)
+		"Could not find/guess checksum in %s: %s", file, checksumType)
 }
 
 // checksumsFromFile will download checksumFile that is a checksum for file
--- a/checksum.go
+++ b/checksum.go
@@ -1,0 +1,164 @@
+package getter
+
+import (
+	"bufio"
+	"crypto/md5"
+	"crypto/sha1"
+	"crypto/sha256"
+	"crypto/sha512"
+	"encoding/hex"
+	"fmt"
+	"hash"
+	"io/ioutil"
+	"net/url"
+	"path/filepath"
+	"strings"
+
+	urlhelper "github.com/hashicorp/go-getter/helper/url"
+)
+
+var checksummers = map[string]func() hash.Hash{
+	"md5":    md5.New,
+	"sha1":   sha1.New,
+	"sha256": sha256.New,
+	"sha512": sha512.New,
+}
+
+func checksumHashAndValue(u *url.URL) (checksumHash hash.Hash, checksumValue []byte, err error) {
+	q := u.Query()
+	v := q.Get("checksum")
+
+	if v == "" {
+		return nil, nil, nil
+	}
+
+	// Determine the checksum hash type
+	checksumType := ""
+	idx := strings.Index(v, ":")
+	if idx > -1 {
+		checksumType = v[:idx]
+	}
+	if fn, found := checksummers[checksumType]; found {
+		checksumHash = fn()
+		// Get the remainder of the value and parse it into bytes
+		checksumValue, err = hex.DecodeString(v[idx+1:])
+		return
+	}
+
+	if checksumType != "file" {
+		return nil, nil, fmt.Errorf(
+			"unsupported checksum type: %s", checksumType)
+	}
+
+	checkSums, err := checksumsFromFile(v[idx+1:], u)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var checksumValueString string
+	for checksumType, checksumValueString = range checkSums {
+		if fn, found := checksummers[checksumType]; found {
+			checksumHash = fn()
+			checksumValue, err = hex.DecodeString(checksumValueString)
+			return
+		}
+	}
+
+	return
+}
+
+func checksumsFromFile(checksumFile string, src *url.URL) (checkSums map[string]string, err error) {
+
+	checksumFileURL, err := urlhelper.Parse(checksumFile)
+	if err != nil {
+		return nil, err
+	}
+
+	f, err := ioutil.TempFile("", "go-getter.checksum")
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close()
+
+	if err = GetFile(f.Name(), checksumFile); err != nil {
+		return nil, fmt.Errorf(
+			"Error downloading checksum file: %s", err)
+	}
+
+	filename := filepath.Base(src.Path)
+	absPath, _ := filepath.Abs(src.Path)
+	relpath, _ := filepath.Rel(filepath.Dir(checksumFileURL.Path), absPath)
+
+	// possible file identifiers:
+	options := []string{
+		filename,       // ubuntu-14.04.1-server-amd64.iso
+		"*" + filename, // *ubuntu-14.04.1-server-amd64.iso  Standard checksum
+		relpath,        // dir/ubuntu-14.04.1-server-amd64.iso
+		"./" + relpath, // ./dir/ubuntu-14.04.1-server-amd64.iso
+		absPath,        // fullpath; set if local
+	}
+
+	rd := bufio.NewReader(f)
+	res := map[string]string{}
+	for {
+
+		line, err := rd.ReadString('\n')
+		if err != nil && line == "" {
+			break
+		}
+		checksumType, checksumValue, filename, ok := parseChecksumLine(checksumFile, line)
+		if !ok {
+			continue
+		}
+		for _, option := range options {
+			// filename matches src ?
+			if filename == option {
+				res[checksumType] = checksumValue
+			}
+		}
+
+	}
+	return
+}
+
+// parseChecksumLine takes a line from a checksum file and returns
+// checksumType, checksumValue and filename
+// For GNU-style entries parseChecksumLine will try to guess checksumType
+// based on checksumFile which should usually contain the checksum type.
+// checksumType will be lowercase.
+//
+// BSD-style checksum:
+//  MD5 (file1) = <checksum>
+//  MD5 (file2) = <checksum>
+//
+// GNU-style:
+//  <checksum>  file1
+//  <checksum> *file2
+func parseChecksumLine(checksumFilename, line string) (checksumType, checksumValue, filename string, ok bool) {
+	parts := strings.Fields(line)
+
+	ok = true
+	switch len(parts) {
+	case 4: // BSD-style
+		if len(parts[1]) <= 2 || parts[1][0] != '(' || parts[1][len(parts[1])-1] != ')' {
+			return "", "", "", false
+		}
+		checksumType = strings.ToLower(parts[0])
+		filename = parts[1][1 : len(parts[1])-1]
+		checksumValue = parts[3]
+	case 2: // GNU-style
+
+		for ctype := range checksummers {
+			// guess checksum type from filename
+			if strings.ContainsAny(checksumFilename, ctype) {
+				checksumType = ctype
+			}
+		}
+		checksumValue = parts[0]
+		filename = parts[1]
+	default:
+		ok = false
+	}
+
+	return
+}
--- a/checksum.go
+++ b/checksum.go
@@ -24,6 +24,27 @@ var checksummers = map[string]func() hash.Hash{
 	"sha512": sha512.New,
 }
 
+// checksumHashAndValue will return checksum based on checksum parameter of u
+// ex:
+//  http://hashicorp.com/terraform?checksum=<checksumType>:<checksumValue>
+//  http://hashicorp.com/terraform?checksum=file:<checksum_url>
+// when checksumming from a file checksumHashAndValue will go get checksum_url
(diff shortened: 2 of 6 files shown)


azr:
I just realised I have some more testing to do; marking as WIP for now.

checksum_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+func u(t *testing.T, in string) *url.URL {
+	u, err := url.Parse(in)
+	if err != nil {
+		t.Fatalf("cannot parse %s: %v", in, err)
+	}
+	return u
+}
+
+func Test_checksumHashAndValue(t *testing.T) {
```
vancluever: Just thinking - I think we should have a full matrix of tests here:

* BSD MD5/SHA
* GNU MD5/SHA

I think in the case of GNU specifically, and both for SHAs, the hash length is also reflective of the algorithm in use, correct? I haven't looked 100%.

checksum.go
```diff
@@ -0,0 +1,213 @@
...
+		"Could not find/guess checksum in %s: %s", file, checksumType)
+}
+
+// checksumsFromFile will download checksumFile that is a checksum for file
+// behind src.
+//
+// checksumsFromFile will try to guess the hashing algorithm based on content
+// of or name of checksum file
+func checksumsFromFile(checksumFile string, src *url.URL) (checkSums map[string]string, err error) {
+
```
vancluever: _Minor_ blank line here should be removed

checksum.go
```diff
@@ -0,0 +1,213 @@
...
+// GNU-style:
+//  <checksum>  file1
+//  <checksum> *file2
+//
+// For GNU-style checksum files; it is very common that the hashing algorithm identifier
+// is in the filename; so the name of every supported hashing algorithm is compared
+// against checksum_url for a match/guess.
+// In case a different hashing algorithm is in the filename of checksum_url
+// it is recommended to explicitly set hashing algorithm instead.
+func checksumHashAndValue(u *url.URL) (checksumHash hash.Hash, checksumValue []byte, err error) {
```
vancluever: This is a pretty complex function and it's probably not a good idea to use naked return values here. Case in point, it is not in use 100% in this function (namely when errors are being returned).

More on naked returns: https://youtu.be/ltqV6pDKZD8?t=1940

checksum.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	// Determine the checksum hash type
+	checksumType := ""
+	idx := len(v)
+	if i := strings.Index(v, ":"); i > -1 {
+		idx = i
+	}
+	checksumType = v[:idx]
+	if fn, found := checksummers[checksumType]; found {
```
vancluever: This was kind of hard to follow. I wonder if this would just be much more beneficial being more explicit here, even if it results in some duplication? Namely, using the old `switch` approach used in `client.go`.

I would probably roll `checksumHashAndValue` into a meta-function that would either return values from `checksumsFromFile` (which would be converted to be standalone) and `checksumsFromQuery` (which would contain most of the code removed from `client.go`).

checksum.go
azr:
```suggestion
```

checksum_test.go
```diff
@@ -0,0 +1,73 @@
...
+
```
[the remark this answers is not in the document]
azr:
```suggestion
```

azr:
Hey @vancluever !
Awesome, thanks for reviewing ! Derp, I totally didn't think of using the lengths of the hashes. I was too focused tried to have pair functionality with packer 🙂.

README.md
```diff
@@ -156,16 +156,32 @@ a checksum for you. Note that checksumming only works for downloading files,
...
 
 ```
 ./foo.txt?checksum=md5:b7d96c89d09d9e204f5fedc4d5d55b21
 ```
 
+```
+./foo.txt?checksum=file:./foo.txt.sha256sum
+```
+ 
+When checksumming from a file, go-getter will go get checksum_url
```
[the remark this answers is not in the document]
azr: Okay, fixed, this should be better, tell me 🙂

README.md
```diff
@@ -156,16 +156,32 @@ a checksum for you. Note that checksumming only works for downloading files,
...
+in a temporary directory and parse the content of the file.
+Content of files are expected to be BSD style or GNU style.
+For GNU-style checksum files; it is very common that the hashing algorithm identifier
```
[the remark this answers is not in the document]
azr: I rewrapped this paragraph to 79, looks better 🙂

azr:
Hey @vancluever, again, thanks for reviewing. I think this is better. I also took some time to add tests for checksumming as there wasn't before.

checksum.go
```diff
@@ -0,0 +1,233 @@
...
+	q := u.Query()
+	v := q.Get("checksum")
+
+	if v == "" {
+		return nil, nil
+	}
+
+	vs := strings.SplitN(v, ":", 2)
+	if len(vs) != 2 {
+		return nil, fmt.Errorf("non explicit checksum type: %s", v)
```
azr: here instead, if the user just did `checksum=<checksum>` we could guess checksum based on length ? Would that be a good idea though ?

azr: I added a commit that does that, easy to revert; but if you think it's a good idea that would avoid a back and forth 🙂

azr:
Okay, there, this should be done.
I also added the possibility to checksum without passing the checksum type in the last commit. I totally can revert it if you think it's a bad idea. 🙂 I did it just because it was really simple to do from where I was and it makes usage simpler as you can now just pass `checksum=<shasum>`

checksum.go
vancluever: Checking on _byte_ length is fine - from what I understand that this is actually how tools like `shasum` and what not check when looking at a format file. Ultimately, since all algos have a static hash length, if it guesses wrong, the hash will fail anyway, and a false negative is better than a false positive. :slightly_smiling_face: :

README.md
```diff
@@ -156,16 +156,40 @@ a checksum for you. Note that checksumming only works for downloading files,
...
+./foo.txt?checksum=b7d96c89d09d9e204f5fedc4d5d55b21
+```
+
+```
+./foo.txt?checksum=file:./foo.txt.sha256sum
+```
+ 
+When checksumming from a file - ex: with `checksum=file:url` - go-getter will
+get the file linked in the URL after `file:` using a new getter
+(`getter.GetFile`). For example, in
```
vancluever: I was expecting this to be the same getter? The below example would indicate this, and especially if this was a remote source the HTTP getter may need to have the same client and headers that are in the getter used to download the target.

README.md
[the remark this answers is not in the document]
vancluever: Looks like this is almost addressed, got a new comment that I will add in the next wave of reviews.

source.go
```diff
@@ -6,24 +6,30 @@ import (
 	"strings"
 )
 
-// SourceDirSubdir takes a source and returns a tuple of the URL without
-// the subdir and the URL with the subdir.
-func SourceDirSubdir(src string) (string, string) {
+// SourceDirSubdir takes a source and returns the URL without
+// the subdir and the subdir.
+func SourceDirSubdir(src string) (url string, subdir string) {
```
vancluever: The prameterized return values here should be removed.

Just looking at the code there does not seem to be any changes introduced that would require a naked return. I'm guessing these were added to save on possible variable declarations - if that's the case, they should be just added as variables either with `var` or initialization operators.

test-fixtures/checksum-file/sha512-bsd.sum
```diff
@@ -0,0 +1 @@
+SHA512 (content.txt) = 060a8cc41c501e41b4537029661090597aeb4366702ac3cae8959f24b2c49005d6bd339833ebbeb481b127ac822d70b937c1637c8d0eaf81b6979d4c1d75d0e1
```
vancluever: Just curious, do we have examples of this in the wild? I can find citations for MD5 but not necessarily SHA.

I don't think we need to worry about blocking these cases if we don't find any though as the implementation is simpler by accounting for these as a result (as the algo is lowercased generally and exceptions would be of low value).

checksum.go
```diff
@@ -0,0 +1,245 @@
...
+
+// checksumsFromFile will download checksumFile that is a checksum for file
+// behind src.
+//
+// checksumsFromFile will try to guess the hashing algorithm based on content
+// of checksum file
+//
+// checksumsFromFile will only return checksums for files that match file
+// behind src
+func checksumFromFile(checksumFile string, src *url.URL) (*fileChecksum, error) {
```
vancluever: Hmm, I think I answered the question that I had from the docs here... we should really try and see how we can ensure that the same getter that is used to fetch the file is used to fetch the checksums. As mentioned, this is important on the HTTP getter where a custom client or additional headers will be used.

checksum.go
```diff
@@ -0,0 +1,245 @@
...
+		os.Remove(tempfile)
+	}()
+
+	if err = GetFile(tempfile, checksumFile); err != nil {
+		return nil, fmt.Errorf(
+			"Error downloading checksum file: %s", err)
+	}
+
+	filename := filepath.Base(src.Path)
+	absPath, _ := filepath.Abs(src.Path)
```
vancluever: Is there a reason why errors are being ignored here? They should probably be returned, while path parsing errors are probably going to be rare mid-request like this, having them is a good idea anyway.

checksum.go
```diff
@@ -0,0 +1,245 @@
...
+		return newChecksumFromType(parts[0], parts[3], filename)
+	case 2:
+		// GNU-style:
+		//  <checksum>  file1
+		//  <checksum> *file2
+		return newChecksumFromValue(parts[0], parts[1])
+	case 0:
+		return nil, nil // empty line
+	default:
+		return newChecksumFromValue(parts[0], "")
```
vancluever: I can't tell if this is covered by tests?

README.md
azr: Gotcha; will make the checksum func a client func !

checksum.go
azr: Okay, trying something 🙂 !

checksum.go
azr: I did add some tests in this file. ex: 
https://github.com/hashicorp/go-getter/blob/487c31a11da44526b399b80b2401c63ce5bc5e1d/get_test.go#L291-L295

Should I also test for weird cases ?

azr: Oh right ! I added a test in 724c0bcbca75a183483d63e5a904891d8e8a12fc

test-fixtures/checksum-file/sha512-bsd.sum
azr: I could not find examples of this in the wild no so I made them up ! Okay removing them 🙂

checksum.go
azr: So, for these 2 cases, the errors are ignored because we could still have a match with `filepath.Base(src.Path)` and then not even need to check against absolute or relative paths.

azr: Ah ! I just read this: 
```
// Pwd is the working directory for detection. If this isn't set, some
// detection may fail. Client will not default pwd to the current
// working directory for security reasons.
```

So it might be worth it to actually not use `filepath.Abs`

source.go
vancluever: Yeah, don't add named returns just because you don't know what they are. This ultimately creates addressable variables for them the function. If it's a documentation issue, add documentation.

test-fixtures/checksum-file/sha512-bsd.sum
vancluever: I was more specifically referring to the `parseChecksumLine` function's behavior of reading hash type from the file, but again, don't change anything - it's fine as is.

source.go
azr: Gotcha, will change this 🙂

get.go
```diff
@@ -69,18 +69,23 @@ func init() {
 	}
 }
 
+// DefaultClient is the default client used
+// for all requests.
+var DefaultClient = &Client{
```
vancluever: This should be changed to a function to effectively replicate the functionality of the previous implementation.

This creates a singleton object, each other call was declaring a new `Client` of its own.

test-fixtures/checksum-file/sha512-bsd.sum
azr: you mean we could just go ahead and just guess the checksum from it's len ?

README.md
vancluever: Hmm, just looking at 487c31a, I think you'd just do better with adding the client as another call parameter.

Everything doesn't necessarily need to be an object, and in this case seeing as you only need it in a few places in checksumFromFile, you could probably just deal with the slight unsighlyness of having an extra parameter versus the risk of making several possibly breaking changes throughout the codebase.

checksum.go
vancluever: Just based on what I mentioned [here](https://github.com/hashicorp/go-getter/pull/136#discussion_r242641807), the client should probably just be passed in as a parameter to save having to apply several breaking changes to the code.

checksum.go
vancluever: I don't think anything here really is a case though for not tracking these errors:

> `filepath.Abs` uses `os.Getwd()` which can error when the syscall fails

System call errors should _definitely_ be logged - we should not be continuing on a broken system

> `filepath.Rel` can fail in some special cases

Again, these special cases are things that we should be tracking an failing on anyway.

My point is that there aren't any cases here where it seems that the `filepath` calls fail due to non-fatal issues processing the path (due to some sort of malformed path). If there was, we would want to type-switch or otherwise check the error for these specific cases so that the rest of them were not blocked from being reported.

> So it might be worth it to actually not use filepath.Abs because it calls os.Getwd

Why is `Getwd` such a problem here? I've seldom seen this call fail, so if it does fail it's probably got a good reason and it's possibly something that is going to keep the rest of go-getter from functioning anyway.

azr: Gotcha, will do this, you are totally right !

README.md
azr: I totally agree with you here; I just took that path to avoid having to have to do something like as this feel not idiomatic to me :
```go
oldSrc := client.Src
client.Src = checksumFile
cliend.Get()
client.Src = oldSrc
//...
```

or

```go
checksumClient := *client // copy client
checksumClient.Src = checksumFile
checksumClient.Get()
//...
```

I think I will go for the second option then 🙂

vancluever: Yeah, I keep forgetting about how `Get` is not paramterized on source and destination. So this is partially my bad, sorry about that!

Maybe instead of the full client then, how about just passing the getters?

azr: No worries, I totally skipped over using getters without any explanation 🙂. I shouldn't have.
Mainly because if I do this I need to re-implement part of what the client does and I thought it was better to just have a breaking change instead. To avoid future issues like "getting a checksum file in my special case doesn't work".

But to get the checksum file I need to:
* guess the getter based on url
* then go-get it

The rest of the client features I could think of are not necessary though :
* checksumming a checksum sounds weird ( this cannot be achieved right now because checksum branching case is in client )
* I have never seen archived/git-stored checksum files
-> checksums are files in general

I will go for passing the getters around then, I'm almost tempted to make a `Getters struct` type 🤔 🙂 

Thanks for your patience !

azr:
Okay @vancluever I think this is better now, putting everything down methods/funcs had me cornered into a rabbit hole so instead I took a step back and put the code back/up in client.Get.

I also made a get func that is not a member of anything and takes all possible parameters. This allowed me to used it to get the checksum file without breaking anything and it's not too bad !

This makes the body of `get` pretty long but I think this is for the best.

Tell me what you think 🙂

azr:
Hello @vancluever, I updated this pull request to have the checksum create a copy of the client to download the checksum file. Like we decided.
I am very glad I split commits by content as this was just a matter of reverting changing path and cherry-picking.

I also marked all convos as resolved for less clogging as this PR got huge. Thanks again for your patiente.

common.go
```diff
@@ -0,0 +1,15 @@
+package getter
+
+import (
+	"io/ioutil"
+)
+
+func tmpFile(dir, pattern string) (string, error) {
```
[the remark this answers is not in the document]
azr: Okay, I simplified this. I hope this is good

azr:
I just rebased this branch after #131 was merged 🙂

common.go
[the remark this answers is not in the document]
vancluever: From what I can see, there's some other implementation details of `tempFile` that were missed. I'll have a look over it again and see if I can just manually correct it.

vancluever:
Okay, as mentioned, the old test `tempFile` function needed fixing up still, because the re-implementation did not preserve the old functionality (returning a path for a static file "foo" within a directory created with `tempDir`).

This has now been corrected (you can see the changes in 6d796ed) and I'm just waiting for Travis to finish before merging.

PR #138, opened 2018-12-13
link in support for the tarDecompressor, so it can be used on .tar files

### LINKED ISSUE #19: Add support for tar archives
Some projects distribute with .zip or .tgz, others have only a .tar (example: https://github.com/AcalephStorage/consul-alerts distributes https://bintray.com/artifact/download/darkcrux/generic/consul-alerts-latest-linux-amd64.tar)

It seems that the go-getter library only supports .tgz or .bz2, but not uncompressed .tar by itself.

ketzacoatl:
resolve #19.

### DIFF
--- a/decompress.go
+++ b/decompress.go
@@ -22,6 +22,7 @@ type Decompressor interface {
 var Decompressors map[string]Decompressor
 
 func init() {
+	tarDecompressor := new(tarDecompressor)
 	tbzDecompressor := new(TarBzip2Decompressor)
 	tgzDecompressor := new(TarGzipDecompressor)
 	txzDecompressor := new(TarXzDecompressor)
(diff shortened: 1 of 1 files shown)


leprechau:
LGTM ... this should work.  I think it makes sense to go ahead and export `tarDecompressor` as `TarDecompressor` for consistency.

ketzacoatl:
@leprechau would that be as simple as renaming https://github.com/hashicorp/go-getter/blob/master/decompress_tar.go#L138-L142 (and the reference in my PR), or is there more to do to export the `TarDecompressor` correctly?

ketzacoatl:
This ought to make it easier to resolve https://github.com/hashicorp/nomad/issues/2443.

schmichael (approved):
Sorry this fell through the cracks. Merging and can export tarDecompressor in a followup. Doing so here would require resolving a merge conflict.

ketzacoatl:
Yay! Thanks @schmichael!

PR #149, opened 2019-01-08
Update S3 to allow setting S3Acceleration config setting.

### LINKED ISSUE #148: S3 URL download failure when forcing s3 protocol
The README states that you can force a protocol by prefixing the URL, such as `s3::` in the case of using the s3 protocol. In my situation the following (edited) artifact URL is valid and works for downloading over HTTPS via go-getter:
```https://some-bucket-name.s3-accelerate.amazonaws.com/packagename/1.0.0/packagename-linux-amd64.tar.gz```

When forcing the protocol to s3, the download fails:
```
go run go-getter/main.go s3::https://some-bucket-name.s3-accelerate.amazonaws.com/packagename/1.0.0/packagename-linux-amd64.tar.gz download
2019/01/07 09:30:15 Error downloading: URL is not a valid S3 URL
exit status 1
```

I would expect that when forcing the protocol the download will still work, however, maybe I am missing something here.

jrasell:
When using S3 with transfer acceleration the AWS config setting
S3UseAccelerate should be set to true. This overrides the
S3ForcePathStyle setting and allows the user to make use of the
bucket transfter acceleration.

Closes #148

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,7 @@ are also supported. If the query parameters are present, these take priority.
   * `aws_access_key_id` - AWS access key.
   * `aws_access_key_secret` - AWS access key secret.
   * `aws_access_token` - AWS access token if this is being used.
+  * `s3_use_accelerate` - Set this to `true` to enable S3 Accelerate feature.
 
 #### Using IAM Instance Profiles with S3
 
--- a/get_s3.go
+++ b/get_s3.go
@@ -6,6 +6,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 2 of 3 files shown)


jrasell:
Hey @tombuildsstuff, any chance in giving this a once over?

bflad:
Hi @jrasell 👋 Thanks for submitting this!

At first glance, it seems we might be able to directly use the S3 Transfer Acceleration enabled hostnames (`BUCKET.s3-accelerate.amazonaws.com` and `BUCKET.s3-accelerate.dualstack.amazonaws.com`) to enable this mode of operation, instead of requiring a separate flag of operation and threading that information through various functions.

In particular:
* The `S3Getter.parseUrl()` function could have separate handling for these hostnames
* The `S3Getter.getAWSConfig()` function could conditionally set `Endpoint` and `S3ForcePathStyle` based on these hostnames

What do you think?

thefallentree:
ping: can we still get this merged?

PR #159, opened 2019-01-29
getfile on windows: when a file cannot be symlinked for privilege reasons, copy it

### LINKED ISSUE #39: local transfer required special privileges on Windows
Windows 7 and more have a very bad handling of symlinks, requiring the UAC dialog even if you're administrator (see http://answers.perforce.com/articles/KB/3472 for more details).
Since using symlink is the default behavior for local transfer, that's create weird situations where downloading from internet can be done without special permission but a local file those.
I think go-getter should better handled this case by default.

This issue is related to nomad issue 1714 (https://github.com/hashicorp/nomad/issues/1714).

azr:
Symlinking allows to optimize the space taken by go-get in case the operation runs locally.
But running the go getter without the symlinking privilege can fail in this case.

This will copy the file instead of symlinking it in case we don't have the priviledge, so that the go-get doesn't fail.

closes #39
related: https://github.com/hashicorp/nomad/issues/3491

### DIFF
--- a/get_file_windows.go
+++ b/get_file_windows.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"syscall"
 )
 
 func (g *FileGetter) Get(dst string, u *url.URL) error {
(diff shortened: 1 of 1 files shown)


azr:
Another fix for this would be to have the go-getter do nothing and find a way to tell the caller that the file is local so that the caller can use it in place.

mitchellh:
This _feels_ totally fine? I'd be fine changing the documentation for `Copy` to "If false, attempts to symlink to speed up local files. If the symlink fails, may attempt to copy."

azr:
I have the same feeling you have, it _feels_ fine. Here we go about the doc update 🙂.

PR #160, opened 2019-01-31
make tmpFile close the newly created file in case a future task need to delete it

azr:
This could cause checksumming from file on windows to error because the go-getter would try to download to that temporary file that is still opened.
This probably worked on unix system as the system considered the process/owner as being the same.


azr:
I wouldn't want to bother to get a review for such a simple bug, so I will merge it. You can still comment here if you think something is wrong and I will for sure make a new PR 🙂

PR #161, opened 2019-01-31
make helper/url.Parse os agnostic

azr:
Hello,
Before this pr, on windows `url.Parse("c:\")` would return `./c:/`.
Now it returns `file://c:/`.

Sometime packer does operations for windows from a unix system therefore we need the Parse func to be os agnostic.
This makes the tests pass on all platform by forcing "C:" prefixed urls to become "file://C:" for example. 

I hope this is all right 🙂


azr:
Oh, but I just realised the go-getter was using a lot of things like `filepath.Abs`. I will close this and just make the change be about the windows fix and not the 'agnostic' part.

PR #162, opened 2019-02-01
Fix tests for windows

azr:
Howdy,
packer runs on windows and I wanted to fix the tests to avoid future regressions !
🙂 

I tried to keep everything as close as possible as before but some behaviour needed to change on windows; see `helper/url/url_windows.go` for more.


checksum.go
```diff
@@ -197,8 +197,15 @@ func (c *Client) checksumFromFile(checksumFile string, src *url.URL) (*fileCheck
 	if err != nil {
 		return nil, err
 	}
-	relpath, err := filepath.Rel(filepath.Dir(checksumFileURL.Path), absPath)
-	if err != nil {
+	checksumFileDir := filepath.Dir(checksumFileURL.Path)
+	relpath, err := filepath.Rel(checksumFileDir, absPath)
+	switch {
+	case err == nil ||
+		err.Error() == "Rel: can't make "+absPath+" relative to "+checksumFileDir:
```
mitchellh: One issue with these comparisons on Windows in particular is that windows error messages are localized. So this will only work for English-language Windows installations.

azr: That one is from Rel: https://github.com/golang/go/blob/03a9f5a192149dc6f443eb0e1cee48cb4dd7e26f/src/path/filepath/path.go#L285

 I reaaaly wish this error had a type though ...

azr:
Yes, travis-ci tests on Linux, I'll make it test on mac.

checksum.go
azr: Do you think there is a better way for this one ?

azr:
Nice, everything is fixed and ci-tested on linux, darwin and windows ! 🙂

azr:
Thanks for your time @/mitchellh  🙂

PR #163, opened 2019-02-05
Various fixes

azr:
This is a bunch of minor fix/improvement after #136.

* passes the cancellation context and progress tracker to a checksum go-getter.
* make sure  HttpGetter.GetFile closes a fd in every case after opening one.
* reset checksum hash before checksumming as it is run before a download to see if the file is already present; this would cause the checksum to error in case the file was only partially present.

### DIFF
--- a/get_http.go
+++ b/get_http.go
@@ -128,7 +128,7 @@ func (g *HttpGetter) Get(dst string, u *url.URL) error {
 	return g.getSubdir(ctx, dst, source, subDir)
 }
 
-func (g *HttpGetter) GetFile(dst string, src *url.URL) (err error) {
+func (g *HttpGetter) GetFile(dst string, src *url.URL) error {
 	ctx := g.Context()
 	if g.Netrc {
 		// Add auth from netrc if we can
@@ -146,12 +146,7 @@ func (g *HttpGetter) GetFile(dst string, src *url.URL) (err error) {
 	if err != nil {
 		return err
 	}
-	defer func() {
-		cerr := f.Close()
-		if err == nil {
-			err = cerr
-		}
-	}()
+	defer f.Close()
 
 	if g.Client == nil {
 		g.Client = httpClient
--- a/get_http.go
+++ b/get_http.go
@@ -128,7 +128,7 @@ func (g *HttpGetter) Get(dst string, u *url.URL) error {
 	return g.getSubdir(ctx, dst, source, subDir)
 }
 
-func (g *HttpGetter) GetFile(dst string, src *url.URL) error {
+func (g *HttpGetter) GetFile(dst string, src *url.URL) (err error) {
 	ctx := g.Context()
 	if g.Netrc {
 		// Add auth from netrc if we can
@@ -146,6 +146,12 @@ func (g *HttpGetter) GetFile(dst string, src *url.URL) error {
 	if err != nil {
 		return err
 	}
+	defer func() {
+		cerr := f.Close()
+		if err == nil {
+			err = cerr
+		}
+	}()
 
 	if g.Client == nil {
 		g.Client = httpClient
@@ -210,9 +216,6 @@ func (g *HttpGetter) GetFile(dst string, src *url.URL) error {
 	n, err := Copy(ctx, f, body)
 	if err == nil && n < resp.ContentLength {
 		err = io.ErrShortWrite
-	}
-	if err1 := f.Close(); err == nil {
-		err = err1
 	}
 	return err
 }
--- a/get_http_test.go
+++ b/get_http_test.go
@@ -1,6 +1,8 @@
 package getter
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -167,12 +169,16 @@ func TestHttpGetter_none(t *testing.T) {
 
 func TestHttpGetter_resume(t *testing.T) {
 	load := []byte(testHttpMetaStr)
+	sha := sha256.New()
+	if n, err := sha.Write(load); n != len(load) || err != nil {
+		t.Fatalf("sha write failed: %d, %s", n, err)
+	}
+	checksum := hex.EncodeToString(sha.Sum(nil))
 	downloadFrom := len(load) / 2
 
 	ln := testHttpServer(t)
 	defer ln.Close()
 
-	g := new(HttpGetter)
 	dst := tempDir(t)
 	defer os.RemoveAll(dst)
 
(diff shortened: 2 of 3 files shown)


get_http_test.go
```diff
@@ -181,28 +187,39 @@ func TestHttpGetter_resume(t *testing.T) {
...
+		RawQuery: "checksum=" + checksum,
+	}
+	t.Logf("url: %s", u.String())
 
-	// Get it!
-	if err := g.GetFile(dst, &u); err != nil {
-		t.Fatalf("should not error: %v", err)
+	// Finish getting it!
+	if err := GetFile(dst, u.String()); err != nil {
+		t.Errorf("finishing download should not error: %v", err)
```
vancluever: Is the test worth continuing here if this fails?

get_http_test.go
```diff
@@ -181,28 +187,39 @@ func TestHttpGetter_resume(t *testing.T) {
...
 
 	b, err := ioutil.ReadFile(dst)
 	if err != nil {
-		t.Fatalf("should not error: %v", err)
+		t.Fatalf("readfile failed: %v", err)
 	}
 
 	if string(b) != string(load) {
-		t.Fatalf("file differs: got:\n%s\n expected:\n%s\n", string(b), string(load))
+		t.Errorf("file differs: got:\n%s\n expected:\n%s\n", string(b), string(load))
```
vancluever: As with the other `Errorf` messages here - is its worth continuing here?

get_http_test.go
```diff
@@ -181,28 +187,39 @@ func TestHttpGetter_resume(t *testing.T) {
...
 	}
 
+	// Get it again
+	if err := GetFile(dst, u.String()); err != nil {
+		t.Errorf("should not error: %v", err)
```
vancluever: As with the other `Errorf` messages here.

As a general rule, `Fatalf` is usually good as if basically indicates that the test case is now "tainted" in a sense in that there's no way we can continue reliably. `Errorf` is fine if we have multiple errors, but how can we be sure that those errors are valid errors in the event of a failure up the line?

get_http_test.go
azr: In my case it was, because the checksum failed here, but then the next checksum passed. ( I just had to `hash.Reset()` to fix )

get_http_test.go
azr: Here, yes too, sam answer as in https://github.com/hashicorp/go-getter/pull/163#discussion_r254781399

get_http_test.go
```diff
@@ -181,28 +187,39 @@ func TestHttpGetter_resume(t *testing.T) {
 	if err != nil {
 		t.Fatalf("create: %v", err)
 	}
-	f.Write(load[:downloadFrom])
-	f.Close()
+	if n, err := f.Write(load[:downloadFrom]); n != downloadFrom || err != nil {
+		t.Fatalf("partia file write failed: %d, %s", n, err)
```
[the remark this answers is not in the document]
azr: Fixed, thanks :)

get_http_test.go
azr: See https://github.com/hashicorp/go-getter/pull/163#discussion_r254781399 also

get_http_test.go
azr: This gave me insightful information as to where something was wrong, but may be now that this issue is fixed, do you think I should rather `Fatal` ?

get_http.go
```diff
@@ -146,6 +146,12 @@ func (g *HttpGetter) GetFile(dst string, src *url.URL) error {
 	if err != nil {
 		return err
 	}
+	defer func() {
```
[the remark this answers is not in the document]
azr: This defer doesn't mask the actual error and will only return the Close error when the main body did not fail.
It was brought by https://github.com/hashicorp/go-getter/commit/d82b370324f4239b505c097a470c46742c536785 and I just tried to keep the same behaviour but also making sure the file is closed on previous errors.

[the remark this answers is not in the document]
azr: I agree that ignoring the close error here is okay and will do this though 🙂. https://github.com/hashicorp/go-getter/commit/d82b370324f4239b505c097a470c46742c536785 did not fix anything.

get_http.go
```diff
@@ -128,7 +128,7 @@ func (g *HttpGetter) Get(dst string, u *url.URL) error {
 	return g.getSubdir(ctx, dst, source, subDir)
 }
 
-func (g *HttpGetter) GetFile(dst string, src *url.URL) error {
+func (g *HttpGetter) GetFile(dst string, src *url.URL) (err error) {
```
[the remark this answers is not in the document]
azr: Yes, removing this, see https://github.com/hashicorp/go-getter/pull/163#discussion_r254785401

get_http.go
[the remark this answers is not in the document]
vancluever: @azr ahh, yeah I see what was happening now, in that it deviates from the standard `if err != nil` pattern, and is actually doing the _reverse_ to populate the return with the close error if no previous error existed. Fair enough!

I still think the alternative is easier to understand, as you can see by the probably common trap I fell in here. In addition, we might not catch _all_ close errors here as the error will not be reported if there was a previous error.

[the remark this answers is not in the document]
azr: Yup, hehe, I totally agree with you 🙂

get_http_test.go
vancluever: I think that maybe the spirit of `Errorf` is useful when you need multiple assertions in a single test.

Consider:

```go
func TestFoo(t *testing.T) {
  x, err := foo("a", "b")
  if err != nil {
    t.Fatalf("error: %s", err)
  }

  if x.A != "a" {
    t.Errorf(`x.A: expected "a", got %q`, x.A)
  }

  if x.B != "B" {
    t.Errorf(`x.B: expected "b", got %q`, x.B)
  }
}
```

In this application, `Errorf` allows you to check both fields in the same test and display failures if _both_ of them fail at the same time, versus having field `A` take priority. However, if there was an error funning `foo`, then more than likely the result in `x` will be invalid and not suitable for giving valid test results (fields could be zero values, or `x` could be `nil`, causing a panic).

In this application, you are not asserting multiple final results, you are mainly testing for a final result of success (no failures). Each failure down the test case ultimately renders the rest of the function unstable, or it be should considered as such. Hence, while for debugging it might have been useful, from a testing standpoint strictly speaking it's not, as an early failure is going to affect everything below it, possibly giving incorrect and/or non-useful error data for operations afterwards.

azr: Ah, yes, 👍 awesome explanation, I'll make them Fatal 🙂

azr: Thanks !!

PR #168, opened 2019-02-21
add a public type for a checksum error

azr:
this will allow to tell if the checksum failed in a get in order to take action, ex: delete the file.

Context:
In packer, an iso is downloaded to `"sha1(file_checksum).iso"`, meaning we must delete the file if the checksum did not match.
If the url/file needs to change, the checksum can match only if we remove the old file.
If the checksum changes, the destination (`"sha1(file_checksum).iso"`)  will change.

### DIFF
--- a/checksum.go
+++ b/checksum.go
@@ -27,6 +27,27 @@ type fileChecksum struct {
 	Filename string
 }
 
+// A ChecksumError is returned when a checksum differs
+type ChecksumError struct {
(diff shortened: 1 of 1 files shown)


azr:
Hello there, I'm going to merge this one as it feels safe and simple enough, if someone wants a change here, please ping me and I will fix it asap.

PR #171, opened 2019-03-04
Adds support for Symlinks in all Tar decompressors

### LINKED ISSUE #60: Support for symlink when untarring an archive
**Operating System:**
Ubuntu 16.04

**Issue:**
When a tarball has symlink inside it, go-getter turns the symlink into zero-length file. One benefit of having go-getter support this would be with respect to Nomad specifically with the artifact stanza https://www.nomadproject.io/docs/job-specification/artifact.html, since some tarball could contain symlink in it.

**Repro:**
 1. Create a test_dir.tar.gz with the following contents (note: test_ls.symlink)
```
vagrant@nomad-server01:/tmp$ tree test_dir/
test_dir/
├── a.txt
└── test_ls.symlink -> /bin/ls
```

 2. On the same directory, execute: `python -m SimpleHTTPServer 8000`
 3. Run `go-getter http://127.0.0.1:8000/test_dir.tar.gz ~/`
 4. Check what go-getter untarred (note: test_ls.symlink is 0 bytes)
```
vagrant@nomad-server01:~$ ls -l ~/test_dir/
total 4
-rw-r--r-- 1 vagrant vagrant 4 May  2 01:19 a.txt
-rwxr-xr-x 1 vagrant vagrant 0 May  2 01:19 test_ls.symlink
```

prologic:
This is useful when using Nomad's Artefact mechanisms which internally
uses `hashicorp/go-getter` to unpack Tar archives that are built from
container root file systems (*which usually contain symlinks for runtime
libraries*).

Fixes #60 and supercedes the very old PR #37

(We use a variant of this patch in our production systems and we need to
get this into upstream as soon as conveniently possible)


prologic:
**NB:** I _deliberately_ did not add Hard link support here as we don't really need it for our use-case and that can come later as a separate PR.

prologic:
🏓

prologic:
> This LGTM ! And tests check that symlinks are written as symlinks ! 👍.
> 
> Too bad for hardlinks but may be in another PR.

Thanks for the review and approval! Yeah we don't really need hardlinks anyway as I know of no root file-system tooling that generate hardlinks instead of symlinks for system/application libraries (`.so`, `.so.1` and such).

Who shall merge this?

prologic:
> Who shall merge this?

nvm :) Thanks!

PR #177, opened 2019-05-08
make Get to return an Operation

azr:
This PR build on top of https://github.com/hashicorp/go-getter/pull/176, to merge in master-v2.

This PR makes the go-getter return an Operation struct telling - for now - where the gotten file or directory is located ( + tests )

PR #176, opened 2019-05-08
Split client and request for a get

azr:
Hello, this PR is a breaking change to merge in the master-v2 branch.

It splits client and request structs, all behaviours are similar.

PR #178, opened 2019-05-08
File inplace

azr:
This PR build on top of #177 & #176, to merge in master-v2.

This PR adds the `Inplace bool` field to a Request allowing to use local files 'inplace' and just setting the `Operation.Dst` to the source file.

This will allow Packer to not do any operation and to use big local files 'inplace'.


get_file.go
```diff
@@ -103,6 +108,11 @@ func (g *FileGetter) GetFile(ctx context.Context, req *Request) error {
 		return err
 	}
 
+	if req.Inplace {
+		req.Dst = path
+		return nil
+	}
+
```
notnoop: Should this be moved above `MkdirAll` call or Dst presence checks?  I'd assume so if we wanted to avoid any operation when running in place.

get_test.go
```diff
@@ -727,3 +727,33 @@ func TestGetFile_inplace(t *testing.T) {
...
+		Getters: map[string]Getter{
+			"file": getter,
+		},
+	}
+
+	// get the file
+	op, err := client.Get(ctx, req)
+	if err == nil {
+		t.Fatalf("err is nil")
+	}
```
notnoop: Can we make a stricter check here?  like checking the type of error or content of error, to confirm it's a checksum error rather than file not found error for example?

get_file.go
azr: Yes this is totally correct ! It's been a while but I wonder why it did this. Moving this up before the Lstat of the dst

get_test.go
azr: Absolutely ! 👍 Done !

get_test.go
```diff
@@ -753,6 +753,10 @@ func TestGetFile_inplace_badChecksum(t *testing.T) {
 	if err == nil {
 		t.Fatalf("err is nil")
 	}
+	_, ok := err.(*ChecksumError)
+	if ok == false {
```
notnoop: It's conventional to condense this if it's not the variables aren't used elsewhere:

```suggestion
	if _, ok := err.(*ChecksumError); !ok {
```

azr:
Hey @notnoop , awesome thanks :D; I'd recommend on merging/reviewing #176, then #177. This will allow to sort of split the concerns and makes for tiny(er) PRs. 
On other thing we can do is review and merge this current PR (#178) which will close 176 and 177 at the same time, the only issue is that it's quite a big scope to chew at once.

azr:
Sorry I'm closing this PR in order to create it agains a v2 branch; just like we are doing for [hcl/v2](https://pkg.go.dev/github.com/hashicorp/hcl/v2?tab=doc), I'll reopen it soon with a link to this one.

azr:
The new PR can be found here: #232

PR #188, opened 2019-06-27
export checksum code

SwampDragons:
Adrien wrote code to allow the go-getter to handle checksums from files, so that Packer didn't need to handle file checksums in its own internal logic. But we missed a "fun" edge case where we were using the checksum information pulled from that file to decide whether or not to _upload_ a file to a vmware esxi machine.

Packer needs to be able to access the actual checksums contained within the file. The only way to do that apart from re-implementing all of the file parsing code from go-getter for this one edge case is to allow the client to call the checksum extraction directly. 

I don't think making this method exported is a particularly bad thing, but I'd appreciate a sanity check... maybe from @vancluever who reviewed the original checksum file code.

FWIW I've tested this change within Packer and it does solve my issue.

### DIFF
--- a/checksum.go
+++ b/checksum.go
@@ -115,7 +115,7 @@ func (c *Client) extractChecksum(u *url.URL) (*fileChecksum, error) {
 
 	switch checksumType {
 	case "file":
-		return c.checksumFromFile(checksumValue, u)
+		return c.ChecksumFromFile(checksumValue, u)
 	default:
 		return newChecksumFromType(checksumType, checksumValue, filepath.Base(u.EscapedPath()))
 	}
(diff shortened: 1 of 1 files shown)


vancluever:
Thanks for referencing the PR where you're using this @SwampDragons! It helps to see how it looks in action.

I don't see an issue in exporting this either - the only thing that I'd like to see is comments adjusted to reflect the new export as well, in addition to `fileChecksum` being exported as well.

SwampDragons:
Awesome.  I can do those.  

Also, re: test failures, I've opened #189 to fix this bitbucket failure.

vancluever:
Beauty! Thanks for taking care of that too!

PR #190, opened 2019-06-30
Use standard name for fixtures dir per Go conventions

radeksimko:
https://golang.org/cmd/go/#hdr-Test_packages

> The go tool will ignore a directory named "testdata", making it available to hold ancillary data needed by the tests.


MONEY756:
It will be very good and interesting👍👍

PR #195, opened 2019-08-13
client add Umask to the client config, use it to Chmod files

langmartin:
This adds a `Client.Umask os.FileMode` and uses it to mask file permissions wherever files are created or permissions are set. This version of the patch is based on v1.1.0, it's to be followed by a pr that uses umask with the gcs getters.


mitchellh:
Just want to note on quick review: using the `syscall` package directly is going to result in build errors cross-platform because the `syscall` package is OS-dependent. This PR for example causes go-getter to no longer build on Windows, so the way that umask values and operations are passed around needs to be abstracted.

get_file_copy.go
```diff
@@ -27,3 +28,32 @@ func Copy(ctx context.Context, dst io.Writer, src io.Reader) (int64, error) {
 		}
 	}))
 }
+
+// CopyReader copies from an io.Reader into a file, using umask to create the dst file
+func CopyReader(dst string, src io.Reader, umask os.FileMode) error {
+	dstF, err := os.OpenFile(dst, os.O_RDWR|os.O_CREATE|os.O_TRUNC, mode(0666, umask))
```
schmichael: Since this accepts a mode, can we make CopyReader take the mode instead of the umask and skip the Chmod we do after CopyReader? Would save a ton of syscalls and filesystem metadata writes on large archives with lots of small files.

client.go
```diff
@@ -66,6 +71,25 @@ type Client struct {
 	Options []ClientOption
 }
 
+func (c *Client) umask() os.FileMode {
+	if c == nil {
```
notnoop: I find it a bit odd handling the receiver, `c`, being nil.  It may mask bugs where callers are operating on a `nil` `Client` rather than a valid reference and getting inconsistent masks.

client.go
```diff
@@ -66,6 +71,25 @@ type Client struct {
...
+		um := syscall.Umask(0)
+		syscall.Umask(um)
```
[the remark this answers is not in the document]
notnoop: > Why are we setting the value we just received

Sadly, this is the most portable way to read umask of process:

> It is impossible to use umask() to fetch a process's umask without at
> the same time changing it.  A second call to umask() would then be
> needed to restore the umask.  The nonatomicity of these two steps
> provides the potential for races in multithreaded programs.
>
> Since Linux 4.7, the umask of any process can be viewed via the Umask
> field of /proc/[pid]/status.  Inspecting this field in
> /proc/self/status allows a process to retrieve its umask without at
> the same time changing it.

http://man7.org/linux/man-pages/man2/umask.2.html

client.go
langmartin: agreed! `get_file_test` tests are written in such a way that the client is nil, and I'm not sure if consumers of this library are counting on the current behavior of `FileGetter` (or the other substructs). It may just be concealing bugs in the callers...

get_file_copy.go
langmartin: Good suggestion, which highlights a quirk of `libc` that is inherited by golang's `os.*`. The file mode can't be used to setuid/gid bits on creation, those bits must be explicitly set in a `chmod` call. This is lightly documented but widespread behavior.
See https://github.com/golang/go/issues/25539

The most future/platform proof way to manage umask is probably to calculate the mode ourselves, although on linux and bsds today its unnecessary duplication, libc will strip setuid/setgid bits on linux, and all special permissions on bsds

langmartin:
See also https://github.com/hashicorp/go-getter/tree/f-umask-v110-2, a much smaller set of changes that prevents setuid simply by avoiding direct calls to `Chroot`. On linux (setuid/gid) and bsd (setuid/gid/sticky) `mkdir` and `open` are already masked for those special permissions, and `os.*` inherits that behavior.

decompress_testing.go
```diff
@@ -47,7 +47,7 @@ func TestDecompressor(t testing.T, d Decompressor, cases []TestDecompressCase) {
 			defer os.RemoveAll(td)
 
 			// Decompress
-			err := d.Decompress(dst, tc.Input, tc.Dir)
+			err := d.Decompress(dst, tc.Input, tc.Dir, 0022)
```
schmichael: It doesn't look like we assert the correct mode is used below. Could you add that?

get_file_copy.go
```diff
@@ -27,3 +28,32 @@ func Copy(ctx context.Context, dst io.Writer, src io.Reader) (int64, error) {
...
+		return err
+	}
+	defer dstF.Close()
+
+	_, err = io.Copy(dstF, src)
+	return err
+}
+
+// copyFile copies a file in chunks from src path to dst path, using umask to create the dst file
+func copyFile(ctx context.Context, dst string, src string, mode os.FileMode) (int64, error) {
```
schmichael: Let's unit test these 2 funcs.

client.go
```diff
@@ -66,6 +70,23 @@ type Client struct {
 	Options []ClientOption
 }
 
+// ProcessUmask contains the process wide umask set at init time. see client_unix.go
+var ProcessUmask os.FileMode = 0022
```
mitchellh: I think I'd prefer if this was unexported. If you want to set the umask for go-getter you should set it on the `Client` directly. Package-variable global configuration is somewhat idiomatic (used a lot in the stdlib) but just always felt like a security risk or race waiting to happen.

mitchellh: Note also: if we unexport it today, then if there are use cases we require it exported in the future we can always export it. But if we don't need it today, we can never really unexport it w/o breaking API guarantees. So I think its best to unexport it.

schmichael:
@jbardin pointed out in a discussion that this breaks backward compatibility for other users of go-getter as well as breaks our `tar --preserve-permissions`-esque behavior.

We could maintain backward compat and keep tar's behavior by defaulting to an empty mask and letting Nomad override it to strip setuid bits.

client.go
langmartin: That makes sense to me.

langmartin:
> We could maintain backward compat and keep tar's behavior by defaulting to an empty mask and letting Nomad override it to strip setuid bits.

That also addresses @mitchellh's concerns in passing by eliminating any dependency on the umask. We can avoid the syscall and the global default.

get_file_copy.go
langmartin: this change is backed out now, we need to use chmod directly to preserve the permissions from the archive

decompress_testing.go
langmartin: This turns out to be an awkward place to check permissions, I've added a separate test for permission preserving and masking.

get_file_copy.go
langmartin: I don't think a direct unit test of these adds much value, they're consumed by all the other unit tested components. Since they're private, I thought separate tests might end up over-specifying them.

client.go
langmartin: Followup here, reading the `syscall.Umask` has been eliminated, our default umask is actually just 0, since the backwards compatibility preserving mode is to run like `tar --preserve-permissions`. So this eliminates the windows compatibility problem and this variable completely.

langmartin:
This PR is superseded by #198, and can be closed.

PR #196, opened 2019-08-15
Add Umask configuration, use it to mask decompress permissions

langmartin:
This is a duplicate of https://github.com/hashicorp/go-getter/pull/195, rebased on an earlier version.


client.go
```diff
@@ -248,7 +266,7 @@ func (c *Client) Get() error {
 	// If we're not downloading a directory, then just download the file
 	// and return.
 	if mode == ClientModeFile {
-		err := g.GetFile(dst, u)
+		err := g.GetFile(dst, u, c.Umask)
```
preetapan: should this be c.umask()

langmartin:
This can be closed, the final version is against the master branch and supercedes this PR

PR #198, opened 2019-08-22
umask final version against master

langmartin:
This PR supersedes https://github.com/hashicorp/go-getter/pull/196 and https://github.com/hashicorp/go-getter/pull/195, which contain the same changes rebased on earlier branches.

Add a `Umask` field to the `Client` struct, and use it to mask the preserved permissions when decompressing archive formats. The default `Umask: 0` preserves all permissions (like the current behavior).

This branch contains additional changes from #195, threading the mask and helper functions through code added since v1.1.0.


shoenig:
Closing this, as https://github.com/hashicorp/go-getter/pull/256 is this but rebased with master

PR #200, opened 2019-09-04
Add replace for apache thrift dependency

NeenuAVarghese:
Work around until https://github.com/census-instrumentation/opencensus-go/issues/1158 is fixed


azr:
This was fixed in #204, Thanks !

PR #202, opened 2019-09-04
Update travis test go version

178inaba:
Go is supported for up to two generations.
Since 1.13 was released, 1.11 will no longer be supported.


178inaba:
@azr Thank you for the review!

I wrote an issue about GCS test failure.
https://github.com/hashicorp/go-getter/issues/210

Do I have to do anything for second validation?

Issue #193, opened 2019-08-07
feature request: reduce dependency size


medyagh:
this is a great lib but I was wondering if there is any room for optimization in terms of dependency size ! after adding go-getter dependency, our binary size increased by 15 mb .

https://github.com/kubernetes/minikube/issues/5013

medyagh:
I suspect you could use just go cloud code for both aws and google cloud and you might not need to import both google coud go and aws for dealing with cloud.

azr:
Hello @medyagh, weird question but the getter's detectors and getters are [defaulted](https://github.com/hashicorp/go-getter/blob/bde455061ffa8f9b2a8a8cf4fedbbdb32dc7b9e3/client_option.go#L24-L31) when not set and that is what I saw in minikube's downloader; [the defaults are used](https://github.com/kubernetes/minikube/blob/a8752f18130445240d81dd3957dbf5184f8d0346/pkg/util/downloader.go#L73-L78).

What if you use instead:

```go
client := &getter.Client{
	Src:      urlWithChecksum,
	Dst:      tmpDst,
	Mode:     getter.ClientModeFile,
	Options:  opts,
	Getter:   /* insert everything except aws/gcp getters */
	Detector: /* insert everything except aws/gcp detectors */
}
```

**I have not checked** but may be the go compiler will do some magic here and remove that code path and therefore remove that weight for you.

That would be nice; otherwise I think that splitting those getter/detectors into their own pkg as part of the v2 of the go-getter would be a great option.

afbjorklund:
@azr:
> **I have not checked** but may be the go compiler will do some magic here and remove that code path and therefore remove that weight for you.

Unfortunately the linker is not smart enough to do such dead-stripping of unused code. Adding the suggested configuration actually makes it **bigger**:

`56088	out/minikube-linux-amd64`
`56096	out/minikube-linux-amd64`

```diff
@@ -69,12 +69,29 @@ func (f DefaultDownloader) CacheMinikubeISOFromURL(url string) error {
 	// Predictable temp destination so that resume can function
 	tmpDst := dst + ".download"
 
+	httpGetter := &getter.HttpGetter{
+		Netrc: true,
+	}
+
 	opts := []getter.ClientOption{getter.WithProgress(DefaultProgressBar)}
 	client := &getter.Client{
 		Src:     urlWithChecksum,
 		Dst:     tmpDst,
 		Mode:    getter.ClientModeFile,
 		Options: opts,
[...]

PR #207, opened 2019-09-10
Add build tags for disabling s3 and gcs

afbjorklund:
-tags "go_getter_nos3 go_getter_nogcs"

Since github.com/aws/aws-sdk-go/aws and
cloud.google.com/go/storage are so big...

For #193


medyagh:
+1

PR #213, opened 2019-10-04
Make testHttpServer work on IPv6-only systems, too.

ubschmidt2:


ubschmidt2:
Travis CI build failure seems unrelated.

PR #214, opened 2019-10-07
convert from Travis CI to Circle CI

cotarg:
This PR converts `hashicorp/go-getter` from Travis to Circle.

PR #215, opened 2019-10-09
update circleci config to save mod cache

cotarg:
Updating the circleci config so it will save the go mod cache in response to: https://github.com/hashicorp/go-getter/pull/214#issuecomment-540083238

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,12 +25,7 @@ jobs:
       - restore_cache:
           keys:
             - go-getter-modcache-v1-{{ checksum "go.mod" }}
-
-      # Save go module cache if the go.mod file has changed
-      - save_cache:
-          key: go-getter-modcache-v1-{{ checksum "go.mod" }}
-          paths:
-            - "/go/pkg/mod"
+            - go-getter-modcache-v1
 
       # check go fmt output because it does not report non-zero when there are fmt changes
       - run:
@@ -51,6 +46,12 @@ jobs:
           path: *TEST_RESULTS_PATH
       - store_artifacts:
           path: *TEST_RESULTS_PATH
+
+      # Save go module cache if the go.mod file has changed
+      - save_cache:
+          key: go-getter-modcache-v1-{{ checksum "go.mod" }}
+          paths:
+            - "/go/pkg/mod"
 
 workflows:
   version: 2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,7 +25,14 @@ jobs:
       - restore_cache:
           keys:
             - go-getter-modcache-v1-{{ checksum "go.mod" }}
-            - go-getter-modcache-v1
+
+      - run: go mod download
+
+      # Save go module cache if the go.mod file has changed
+      - save_cache:
+          key: go-getter-modcache-v1-{{ checksum "go.mod" }}
+          paths:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -25,12 +25,7 @@ jobs:
       - restore_cache:
           keys:
             - go-getter-modcache-v1-{{ checksum "go.mod" }}
-
-      # Save go module cache if the go.mod file has changed
-      - save_cache:
-          key: go-getter-modcache-v1-{{ checksum "go.mod" }}
-          paths:
-            - "/go/pkg/mod"
+            - go-getter-modcache-v1
```
alvin-huang: Curious what this for here. This would never match since the caches are saved with `go-getter-modcache-v1-{{ checksum "go.mod" }}` so the key would need to be `go-getter-modcache-v1-` if you meant to match when there is no `go.mod` file to checksum.

.circleci/config.yml
```diff
@@ -52,6 +47,12 @@ jobs:
       - store_artifacts:
           path: *TEST_RESULTS_PATH
 
+      # Save go module cache if the go.mod file has changed
+      - save_cache:
```
alvin-huang: Generally I like to keep this as close to the `restore_cache` as possible so it's easy to see:

1. What cache is restored
2. If the cache isn't there what is run to generate the cache
3. What paths are saved in the cache

The way you have it also works since go fmt downloads dependencies so I will leave it up to you.

.circleci/config.yml
alvin-huang: You probably don't need this at all.

cotarg: I added it per a comment Aaron made, but I'm happy enough to take it out as well... Aaron's suggestion was that it would allow you to get a previous cache if one existed but the checksum didn't match. My assumption there was that you pick up a little efficiency because when it gets to running `go mod` as part of the tests it'll only have to grab packages that aren't already in the saved cache.

I could go either way on it since I don't think the savings is that great and looking at it in the YAML is actually a smidge confusing if you don't know that circle allows the fallback there.

...nevermind, just talked myself around to taking it out based on that. :D

.circleci/config.yml
cotarg: So in this case would you just leave it where it had been before but reinsert the `run: go mod` line? 

By which I mean how it was done here: https://github.com/hashicorp/go-memdb/blob/9854a2567d0b382dfb6f632cc947f887255e331a/.circleci/config.yml#L21-L32

cotarg: I ask because I like the logical ordering of keeping the caching concerns together so it's easily readable/understandable.

alvin-huang: Yeah for me I would just do the explicit module download (`go mod download`) and then save it right after. All those go commands after will still try to download the modules first but since it finds it in the modcache it's very fast to just use what it pulled down in a previous `run` step and move on. 

Sorry to push you back and forth on this!

cotarg: :) It's an opportunity for me to think about it and test my opinions to see if I should keep them! I'll take both of these PRs to that format then. Ultimately I think the small efficiency that might be gained here is really just passing around the same bit of margin in different ways.

alvin-huang:
Looks great!

Issue #152, opened 2019-01-09
S3 go getter doesn't work in CodeBuild


sdmichelini:
Issue that was fixed in the aws provider(https://github.com/terraform-providers/terraform-provider-aws/pull/1425/files) does not seem to be fixed here. The issue arises when you true to import a module from s3 while you are running inside CodeBuild.

m13t:
I can confirm this also does not work from within an ECS container. It seems the code is explicitly looking for EC2 instance metadata URLs and does not check for `AWS_EXECUTION_ENV=AWS_ECS_EC2` environment variables to ascertain whether to get credentials from `AWS_CONTAINER_CREDENTIALS_RELATIVE_URI` instead.

solarmosaic-kflorence:
It is terribly confusing that all other places in Terraform work, and this one fails. Took me awhile to figure out what the root issue was. Please prioritize this fix!

solarmosaic-kflorence:
Here is a workaround, in accordance with https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-iam-roles.html (it can be sourced within a CodeBuild buildspec, for example in the `pre_build` phase).

```
#!/usr/bin/env bash
AWS_ACCESS_KEY_ID=$(curl --silent http://169.254.170.2:80$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI | jq -r '.AccessKeyId')
AWS_SECRET_ACCESS_KEY=$(curl --silent http://169.254.170.2:80$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI | jq -r '.SecretAccessKey')
AWS_SESSION_TOKEN=$(curl --silent http://169.254.170.2:80$AWS_CONTAINER_CREDENTIALS_RELATIVE_URI | jq -r '.Token')
export AWS_ACCESS_KEY_ID
export AWS_SECRET_ACCESS_KEY
export AWS_SESSION_TOKEN
```

Here is the corresponding ticket in terraform: https://github.com/hashicorp/terraform/issues/20439

solarmosaic-kflorence:
@azr I believe this has been resolved by #218

Issue #22, opened 2016-04-18
S3: Support for getting credentials from EC2 meta-data service when using IAM roles.


c4milo:
I can see go-getter using the EC2RoleProvider but I couldn't find where it was calling `creds.Get()` to download the actual credentials, from AWS metadata service. I ran across this issue while trying to use Nomad with a job using artifacts located in S3.

I also manually verified that retrieving credentials using CURL was possible in order to rule out IAM configuration issues.

mitchellh:
I believe this is fixed with this: https://github.com/hashicorp/go-getter/commit/3142ddc1d627a166970ddd301bc09cb510c74edc

We weren't configured the EC2RoleProvider properly.

PR #218, opened 2019-10-14
Use default AWS credential handling under normal circumstances

### LINKED ISSUE #191: AWS credentials not handled correctly
There are various issues with the AWS credential handling when downloading from S3. #185 #157 #152 #22 https://github.com/gruntwork-io/terragrunt/pull/775

I'm less familiar with the AWS SDK for Go, but in my experience with those for JavaScript and Python, the credential handling here [1] looks like it can be made simpler and more robust by using the SDK as documented [2].

The particulars about communicating with the EC2 (or ECS) metadata API, reading the shared config or credentials file, choosing a region, etc., are all handled by the SDK's default credential provider chain. The explicit provider chain constructed in `getAWSConfig` lacks support for various edge cases.

Would you be open to a PR changing the way AWS sessions are created?

[1] https://github.com/hashicorp/go-getter/blob/da0323b9545e4b9d821e9c7d8ed4429e41d997ed/get_s3.go#L182-L217
[2] "Configuring Credentials" section on https://docs.aws.amazon.com/sdk-for-go/api/

jbergknoff-rival:
Fixes #191 (and #185, #157, #152, #22)

This PR changes go-getter to default to the AWS SDK's built in credential resolution. This makes it possible for go-getter to download from S3 using an assumed role (e.g. when setting environment variables `AWS_SDK_LOAD_CONFIG=true AWS_PROFILE=...`, go-getter will assume that role before attempting any operations). It should also add support for various other edge cases relating to AWS credentials.

The one case where go-getter will continue to construct its own credential chain is the one with an override for the instance metadata URL. I'm not familiar with the use case of overriding the instance metadata URL, so I left that code as it was.

I'd like to add a test to protect against regressions in assume-role behavior, but I'd appreciate some guidance there. I see the hardcoded read-only IAM access key/secret. Do those credentials also have the ability to assume some role?

### DIFF
--- a/get_s3.go
+++ b/get_s3.go
@@ -181,20 +181,15 @@ func (g *S3Getter) getObject(ctx context.Context, client *s3.S3, dst, bucket, ke
 
 func (g *S3Getter) getAWSConfig(region string, url *url.URL, creds *credentials.Credentials) *aws.Config {
 	conf := &aws.Config{}
-	if creds == nil {
-		// Grab the metadata URL
-		metadataURL := os.Getenv("AWS_METADATA_URL")
-		if metadataURL == "" {
-			metadataURL = "http://169.254.169.254:80/latest"
-		}
-
+	metadataURLOverride := os.Getenv("AWS_METADATA_URL")
+	if creds == nil && metadataURLOverride != "" {
 		creds = credentials.NewChainCredentials(
 			[]credentials.Provider{
 				&credentials.EnvProvider{},
 				&credentials.SharedCredentialsProvider{Filename: "", Profile: ""},
 				&ec2rolecreds.EC2RoleProvider{
 					Client: ec2metadata.New(session.New(&aws.Config{
-						Endpoint: aws.String(metadataURL),
+						Endpoint: aws.String(metadataURLOverride),
 					})),
 				},
 			})
@@ -213,7 +208,7 @@ func (g *S3Getter) getAWSConfig(region string, url *url.URL, creds *credentials.
 		conf.Region = aws.String(region)
 	}
 
-	return conf
+	return conf.WithCredentialsChainVerboseErrors(true)
 }
 
 func (g *S3Getter) parseUrl(u *url.URL) (region, bucket, path, version string, creds *credentials.Credentials, err error) {


get_s3.go
```diff
@@ -213,7 +208,7 @@ func (g *S3Getter) getAWSConfig(region string, url *url.URL, creds *credentials.
 		conf.Region = aws.String(region)
 	}
 
-	return conf
+	return conf.WithCredentialsChainVerboseErrors(true)
```
jbergknoff-rival: This is just a useful configuration change to help debugging credentials issues.

**Before:**
```
Copying configuration from "s3::https://s3-us-west-2.amazonaws.com/..."...

Error: Failed to download module

Could not download module "root"
(-from-module="s3::https://s3-us-west-2.amazonaws.com/...":1)
source code from
"s3::https://s3-us-west-2.amazonaws.com/...
NoCredentialProviders: no valid providers in chain. Deprecated.
    For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```

**After:**
```
Copying configuration from "s3::https://s3-us-west-2.amazonaws.com/..."...

Error: Failed to download module

Could not download module "root"
(-from-module="s3::https://s3-us-west-2.amazonaws.com/...":1)
source code from
"s3::https://s3-us-west-2.amazonaws.com/...":
NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the
environment.
SharedCredsLoad: failed to load profile, <profile name>.
EC2RoleRequestError: no EC2 instance role found
caused by: RequestError: send request failed
caused by: Get
http://169.254.169.254/latest/meta-data/iam/security-credentials/: dial tcp
169.254.169.254:80: connect: no route to host
```

bashims: This looks good to me and obsoletes my patch. We will test it under ECS to make sure. Thanks!

NinaDenisova:
Hi @jbergknoff-rival I also ran into the issue of not being able to access s3 buckets when it belongs to an originalizational unit. I tested your PR and it does solve the issue for us as well. If I understand it correctly, the next step would be to sign the SLA so that the PR can be merged. I was wondering if you are planning to sign it in the near future. 
Thank you!

jbergknoff-rival:
Oh, thanks for pointing that out, @NinaDenisova . I actually have signed the CLA, and when I visit https://cla.hashicorp.com/hashicorp/go-getter?pullRequest=218, it shows me

image

and the form at the bottom is grayed out, with my details filled in. Maybe there's a bug on the status check.

Either way, I don't think that was the primary blocker on this PR. It just hasn't received any attention from maintainers.

NinaDenisova:
Thank you @jbergknoff-rival for the quick reply! 
@hashicorp-cla Could you please clarify the status of CLA for this PR? 
@hashicorp-cloud I was wondering if anybody has already checked this PR as it's been open for a while now. The changes are not extensive and enable letting AWS take care of credentials provider chain rather than creating the chain manually for the case when credentials are not supplied. This addresses a few issues stated in the PR description (Fixes #191 (and #185, #157, #152, #22)). 
Thank you, everybody.

azr:
Hello there, about the CLA; you need to make sure the email you signed the CLA with is the same as the email you have in your commit, otherwise we cannot merge this.

jbergknoff-rival:
Thanks for the clarification. Now I understand what's going on. At the time I signed the CLA, the email address of the commit was associated with this GitHub account. The company has since become defunct, and I disassociated this GitHub account from that email address (which is also defunct). If I get a chance, I'll try rewriting the commit history on the PR branch.

azr (approved):
LGTM ( just waiting for CLA )

jbergknoff:
:+1: I rewrote the commit to use the appropriate email address and now the CLA check is there. I also rebased from the current master branch.

azr:
Awesome !

apeeters:
How does this relate to / duplicates  https://github.com/hashicorp/go-getter/pull/261?

Issue #225, opened 2019-11-28
Intermittent Git pathspec error when passing depth with ref


AliAllomani:
Intermittent Git pathspec error when passing depth with ref, seems the files are being downloaded but it throws the error afterwards

```
# go-getter "git::https://github.com/hashicorp/packer.git?depth=1" test2
2019/11/28 09:27:15 success!

# du -sh test2
114M	test2


#  go-getter "git::https://github.com/hashicorp/packer.git?depth=1&ref=fix_7677" test4
2019/11/28 09:29:12 Error downloading: error downloading 'https://github.com/hashicorp/packer.git?depth=1&ref=fix_7677': /usr/bin/git exited with 1: error: pathspec 'fix_7677' did not match any file(s) known to git.

# du -sh test4
114M	test4

# go-getter "git::https://github.com/hashicorp/packer.git?ref=fix_7677" test5
2019/11/28 09:30:25 success!

# du -sh test5
138M	test5

# go-getter "git::https://github.com/hashicorp/packer.git?depth=1&ref=hcl2" test6
2019/11/28 09:31:20 Error downloading: error downloading 'https://github.com/hashicorp/packer.git?depth=1&ref=hcl2': /usr/bin/git exited with 1: error: pathspec 'hcl2' did not match any file(s) known to git.

# du -sh test6
114M	test6

# go-getter "git::https://github.com/hashicorp/packer.git?ref=hcl2" test7
2019/11/28 09:32:14 success!

# du -sh test7
136M	test7
```

mhvelplund:
It seems random. It works for some repos and not for others. I checked that I could manually get the same ref using `git clone --depth 1 --branch v0.0.1 git@github.com:my/module.git` and it worked fine ....

PR #226, opened 2019-11-28
fix error when both ref and depth specified

AliAllomani:
Related to Issue #225 

- g.clone doesn't define which branch to clone, so when depth is specified it's only clone master branch
- after that it tries to checkout to the branch with g.checkout(dst, ref)
- since the defined branch hasn't been cloned, it fail with the error message "error: pathspec 'XXX' did not match any file(s) known to git"

### DIFF
--- a/get_git.go
+++ b/get_git.go
@@ -114,7 +114,7 @@ func (g *GitGetter) Get(dst string, u *url.URL) error {
 	if err == nil {
 		err = g.update(ctx, dst, sshKeyFile, ref, depth)
 	} else {
-		err = g.clone(ctx, dst, sshKeyFile, u, depth)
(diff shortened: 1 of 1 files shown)


sebglon:
Hi,
How we can help to fix/merge this PR?
We have the same bug with a very big repository and we want to use the depth.

Regards

sebglon:
We can probably use `git clone --no-checkout --depth 1` to import the git index
And after that use the git checkout already defined on the update func.

It seems better

SergeyMuha:
Waiting for the fix

zachwhaley:
Related #266

william-salt-cko:
@azr sorry to @ you. I think you are one of the main contributors/maintainers of the package, could you perhaps chime in on this, or the associated PRs?

PR #228, opened 2020-01-02
Handle Github query strings with slashes

johananl:
Github branch and tag names can be referenced using the following
format:

    github.com/hashicorp/foo?ref=branch-name

When the branch or tag name contains a slash (which is allowed by Git
as well as Github), splitting the source string by slashes yields an
incorrect result.


andphe:
This would be good, if you have all your terraform in the same repository along with code for other tools (like me) then you may want to have tags like terraform/vN.N.N

johananl:
This has been open with zero traction for almost 3 years. I'm closing the PR.

PR #230, opened 2020-01-28
Split req and client

azr:
This PR replaces #176.

This branch split request arguments from the client into a new request struct.

This is the groundwork to allow the go-getter to use file inplace or tell what operation happened.


client_option.go
```diff
@@ -1,22 +1,10 @@
...
-		c.Ctx = context.Background()
-	}
-	c.Options = opts
-	for _, opt := range opts {
-		err := opt(c)
-		if err != nil {
-			return err
-		}
-	}
+func (c *Client) Configure() error {
```
notnoop: Should this remain a public method?  How do consumers expect to use it?

client_option.go
```diff
@@ -1,22 +1,10 @@
 package getter
 
-import "context"
-
 // A ClientOption allows to configure a client
 type ClientOption func(*Client) error
```
notnoop: I think this is unused now

checksum.go
```diff
@@ -190,7 +191,7 @@ func newChecksumFromValue(checksumValue, filename string) (*FileChecksum, error)
 //
 // ChecksumFromFile will only return checksums for files that match file
 // behind src
-func (c *Client) ChecksumFromFile(checksumFile string, src *url.URL) (*FileChecksum, error) {
+func (c *Client) ChecksumFromFile(ctx context.Context, checksumFile string, src *url.URL) (*FileChecksum, error) {
```
notnoop: Not sure if I follow this function well.  A couple of questions:
* does this function need to be public?  The rest are private ones
* `checksumFile` here is really a go-getter-able address and not a simple path to a local file?  Maybe rename it to `checksumURL` ?
* `src` here isn't really a source for anything relevant to function?  Maybe use `u` like `extractChecksum`?

checksum.go
```diff
@@ -93,7 +94,7 @@ func (c *FileChecksum) checksum(source string) error {
 //  <checksum> *file2
 //
 // see parseChecksumLine for more detail on checksum file parsing
-func (c *Client) extractChecksum(u *url.URL) (*FileChecksum, error) {
+func (c *Client) extractChecksum(ctx context.Context, u *url.URL) (*FileChecksum, error) {
```
notnoop: Not in this change, but I'd stay away from the temp directory implementation details in the function comments.

request.go
```diff
@@ -0,0 +1,42 @@
...
+	// Pwd is the working directory for detection. If this isn't set, some
+	// detection may fail. Client will not default pwd to the current
+	// working directory for security reasons.
+	Src string
+	Dst string
+	Pwd string
+
+	// Mode is the method of download the client will use. See ClientMode
+	// for documentation.
+	Mode ClientMode
```
notnoop: Should we rename `ClientMode` to `RequestMode`?  There is no longer a single mode used by client anymore?

get.go
```diff
@@ -31,16 +32,16 @@ type Getter interface {
...
 	// GetFile downloads the give URL into the given path. The URL must
 	// reference a single file. If possible, the Getter should check if
 	// the remote end contains the same file and no-op this operation.
-	GetFile(string, *url.URL) error
+	GetFile(context.Context, *Request) error
 
 	// ClientMode returns the mode based on the given URL. This is used to
 	// allow clients to let the getters decide which mode to use.
-	ClientMode(*url.URL) (ClientMode, error)
+	ClientMode(context.Context, *url.URL) (ClientMode, error)
```
notnoop: This function seems more `InferURLMode` or `ModeForURL` ?  Also, consider renaming `ClientMode` to `RequestMode`?

notnoop: I see this is addressed in the follow up PRs: https://github.com/hashicorp/go-getter/pull/232/commits/3d28ee4ed4f05246a41a97cc96b3a351e6125aa3 .  Ignore this - thanks!

client_option.go
azr: You are right; making it private for now. It's an artifact of the past 🙂.

client_option.go
azr: Yes; removing it !

checksum.go
azr: * I originally made this private but Packer needs it, and the context can be found here: https://github.com/hashicorp/go-getter/pull/188 , TLDR: we checksum a local file to see wether it needs to be uploaded somewhere.
* correct. I agree, changing this to `checksumURL`
* src is the src of the file being checksummed, checksums in files have a filename attached to it and this filename should match the url being downloaded. The fact that it is an url make it a little complex and only the Path part was used, changing it to a string; also I will rename it to checksummedURL, I hope this makes better sense; tell me what you think.

checksum.go
azr: Because the implementation could change ? Do you want me to change this ?

azr:
Thanks for reviewing @notnoop 🙏 !! Pong !

checksum.go
azr: ( Do you have thoughts on number 3 ? )

notnoop: `checksummedURL` makes sense here - Thanks!  Thanks for the context.

checksum.go
notnoop: don't need to change it, more of a general comment -- it's better to document the function use case or specification in a way that users understand it without necessary fully understand the implementation.  Also, when implementation change, comments are very likely to be out of sync.

checksum.go
```diff
@@ -183,15 +184,15 @@ func newChecksumFromValue(checksumValue, filename string) (*FileChecksum, error)
 	return c, nil
 }
 
-// ChecksumFromFile will return all the FileChecksums found in file
+// ChecksumFromFile will return the first FileChecksums found in file.
 //
-// ChecksumFromFile will try to guess the hashing algorithm based on content
-// of checksum file
+// ChecksumFromFile will try to guess the hashing algorithm based on the
+// content of a checksum file.
```
notnoop: My English is failing me here but a tad vague, it's not immediately obvious which `file` this comment we refer to - maybe the following:

```
// ChecksumFromFile will return the first file checksum found in `checksumURL` file
// that correspond to the `checksummedURL` path.
//
// ChecksumFromFile will infer the hashing algorithm based on the checksumURL file content.
```

checksum.go
azr: Got it, super true, will try to keep that in mind ! Thanks 🙂

checksum.go
```diff
@@ -183,15 +184,15 @@ func newChecksumFromValue(checksumValue, filename string) (*FileChecksum, error)
...
 //
-// ChecksumFromFile will only return checksums for files that match file
-// behind src
-func (c *Client) ChecksumFromFile(checksumFile string, src *url.URL) (*FileChecksum, error) {
-	checksumFileURL, err := urlhelper.Parse(checksumFile)
+// ChecksumFromFile will only return checksums for files that match
+// checksummedURL, which is the object being checksummed.
+func (c *Client) ChecksumFromFile(ctx context.Context, checksumURL, checksummedURL string) (*FileChecksum, error) {
```
notnoop: actually, one more nitpick - now that we pass `u.Path`, `checksummedURL` should be `checksummedPath`?

checksum.go
azr: Oh super good, changing this too.

azr:
Many thanks for the review @notnoop. Given the other positive review from #232; this can finally be merged !!!!

checksum.go
azr: Ah I didn't see this one sorry, you are correct. I will make a new PR changing this.

azr: Ah I pushed it directly to `v2` 🙂

PR #231, opened 2020-01-28
make Get return GetResult

azr:
This pr is based upon #230, merging it will close #230; but I split them in order to allow to review the changes in smaller contextualised chunks. ( which I recommend )

The features of this PR allow to tell what operation was done by the go getter; this will allow the go-getter to 'make a choice' based on a set of arguments.

For example:
* Decide of a random temporary folder/file in which to put the wanted object and then tell us where it was put.
* just tell that this file was not being copied but simply referenced because it is already in the drive ( this one is usefull to Packer )

This PR is reopening #177 but to merge it on the `v2` branch.

### DIFF
--- a/operation.go
+++ b/operation.go
@@ -1,0 +1,7 @@
+package getter
+
+// Operation is the result of a get
+type Operation struct {
+	// Local destination of the gotten object.
+	Dst string
+}
--- a/operation.go
+++ b/operation.go
@@ -1,7 +1,0 @@
-package getter
-
-// Operation is the result of a get
-type Operation struct {
-	// Local destination of the gotten object.
-	Dst string
-}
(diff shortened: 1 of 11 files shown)


operation.go
```diff
@@ -0,0 +1,7 @@
+package getter
+
+// Operation is the result of a get
+type Operation struct {
```
notnoop: This feels more like a `GetResult` or `GetOutput`, as `Get` returns it only after the Get operation completes.  I would have expected `Operation` to represent the underlying operation to download or upload files, and expose more metadata with some asynchorinisty/cancellations/progress hooks.  Some familiar APIs that use Operations are (Google API [1], AWS SDK [2]).

[1] https://godoc.org/google.golang.org/api/compute/v1#Operation
[2] https://docs.aws.amazon.com/sdk-for-go/api/aws/request/#Operation

azr: Thanks for reviewing 🙏. Oh that is a super valid statement. Changing this to `GetResult` !

azr: I also removed the operation file and put `GetResult` in client.go.

azr:
Okay I rebased this change on v2. It's easier to read now

azr:
Thanks @notnoop :D :D

PR #239, opened 2020-03-17
Retrieve checksum from subfolder/filename

sylviamoss:
This PR adds the implementation to retrieve a checksum from a file based on a subfolder + filename. For example: 
Iso [checksum file](http://archive.ubuntu.com/ubuntu/dists/bionic-updates/main/installer-amd64/current/images/SHA1SUMS) contains both below, differing only by the subfolder.
```
a264b6b009dfaa16286fdfd046a156a43587333b  ./hwe-netboot/mini.iso
27c39bac2cf4640c00cacfc8982b0ba39e7b7f96  ./netboot/mini.iso
```
The checksummed path is http://archive.ubuntu.com/ubuntu/dists/bionic-updates/main/installer-amd64/current/images/netboot/mini.iso
We then verify wich checksum filename is contained by the checksummed path if none of the previous options was valid. 

This is motivated by issue https://github.com/hashicorp/packer/issues/8774

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 vendor
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 vendor
-.idea
--- a/checksum.go
+++ b/checksum.go
@@ -216,22 +216,12 @@ func (c *Client) ChecksumFromFile(ctx context.Context, checksumURL, checksummedP
 			"Error downloading checksum file: %s", err)
 	}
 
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -1 +1,2 @@
 vendor
+.idea
```
azr: this is specific to your dev env and I think it belongs in you root gitignore.
```suggestion
```

I have this in my $HOME/.gitconfig
```gitconfig
[core]
	excludesfile = /Users/azr/.gitignore
```

sylviamoss: oh thanks for the tip!!

PR #240, opened 2020-03-26
[WIP] Add SmbGetter

sylviamoss:
This PRs adds a SmbGetter and a SmbDetector to download files from shared folders using SMB protocol. 
On Unix, the SMB getter will try to get files from samba using the [smbclient](https://www.samba.org/samba/docs/current/man-html/smbclient.1.html) when a URL is prefixed with `smb://`, for example `smb://foo/bar/dir`. 
On Windows, the SMB getter will try to get files using the windows file system when the URL is prefixed with `smb://`, `//`, or `\\`. This will be done by using the existing FileGetter if it's available. 
The SmbDetector will make windows src path with `//` or `\\` prefixes a `smb://` URL.

This PR also adds the necessary config for running the tests. This includes:
- A new step to the CircleCi that starts two containers, one for the smb server and the other for the go-getter (client), and run the tests inside the go-getter container;
- Dockerfile and docker-compose files to configure both containers and add shared folders with mock files and directories used on `go_smb_test.go` and `client_test.go`;
- Makefile to help running the containers and tests locally;
- Readme information about the tests and the SMB protocol.

This is motivated by issues https://github.com/hashicorp/packer/issues/8729 and https://github.com/hashicorp/packer/issues/8552 caused by 
https://github.com/hashicorp/packer/pull/7627


SwampDragons:
@schmichael any chance we can get a Nomad review on this?

.circleci/config.yml
```diff
@@ -54,8 +54,40 @@ jobs:
...
+            docker-compose up -d
+
+      - run:
+          name: wait for containers to start
+          command: sleep 60
+
+      - run:
+          name: run smb getter tests
+          command: |
+            docker exec -it samba bash -c "echo 'Hello' > /mnt/file.txt && mkdir -p /mnt/subdir  && echo 'Hello' > /mnt/subdir/file.txt"
```
sylviamoss, line 85: I couldn't create the files in the Dockerfile, this somehow was not working on CircleCi so I had to create them later.

README.md
```diff
@@ -360,3 +361,24 @@ In order to access to GCS, authentication credentials should be provided. More i
 #### GCS Testing
 
 The tests for `get_gcs.go` require you to have GCP credentials set in your environment.  These credentials can have any level of permissions to any project, they just need to exist.  This means setting `GOOGLE_APPLICATION_CREDENTIALS="~/path/to/credentials.json"` or `GOOGLE_CREDENTIALS="{stringified-credentials-json}"`.  Due to this configuration, `get_gcs_test.go` will fail for external contributors in CircleCI.
+
+### SMB (smb)
+
+To access samba shared folder it's necessary to have a [smbclient](https://www.samba.org/samba/docs/current/man-html/smbclient.1.html) installed or a local mount of the shared folder.
```
azr, line 367: I think it should be a good idea to explain what happens on Windows here.

get_smb.go
```diff
@@ -0,0 +1,320 @@
...
+	// download everything that's inside the directory (files and subdirectories)
+	smbclientCmd := baseCmd + " --command 'prompt OFF;recurse ON; mget *'"
+
+	if req.Dst != "" {
+		_, err := os.Lstat(req.Dst)
+		if err != nil {
+			if os.IsNotExist(err) {
+				// Create destination folder if it doesn't exists
+				if err := os.MkdirAll(req.Dst, 0755); err != nil {
+					return fmt.Errorf("failed to creat destination path: %s", err.Error())
```
SwampDragons, line 146:
```suggestion
					return fmt.Errorf("failed to create destination path: %s", err.Error())
```

get_smb.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	// download everything that's inside the directory (files and subdirectories)
+	smbclientCmd := baseCmd + " --command 'prompt OFF;recurse ON; mget *'"
+
+	if req.Dst != "" {
+		_, err := os.Lstat(req.Dst)
+		if err != nil {
+			if os.IsNotExist(err) {
+				// Create destination folder if it doesn't exists
```
SwampDragons, line 144:
```suggestion
				// Create destination folder if it doesn't exist
```

get_smb.go
```diff
@@ -0,0 +1,320 @@
...
+	_, err = runSmbClientCommand(smbclientCmd, req.Dst)
+	return err
+}
+
+func (g *SmbGetter) GetFile(ctx context.Context, req *Request) error {
+	if req.u.Host == "" || req.u.Path == "" {
+		return fmt.Errorf(basePathError)
+	}
+
+	// If dst folder doesn't exists, we need to remove the created on later in case of failures
```
SwampDragons, line 163:
```suggestion
	// If dst folder doesn't exist, we need to remove the created one later in case of failures
```

get_smb.go
```diff
@@ -0,0 +1,320 @@
...
+		return fmt.Errorf("%s source path must be a file", file)
+	}
+
+	// download file
+	smbclientCmd := baseCmd + " --command " + fmt.Sprintf("'get %s'", file)
+	if req.Dst != "" {
+		_, err := os.Lstat(req.Dst)
+		if err != nil {
+			if os.IsNotExist(err) {
+				// Create destination folder if it doesn't exists
```
SwampDragons, line 231:
```suggestion
				// Create destination folder if it doesn't exist
```

get_smb_test.go
```diff
@@ -0,0 +1,410 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestSmbGetter_impl(t *testing.T) {
+	var _ Getter = new(SmbGetter)
+}
+
+func TestSmbGetter_Get(t *testing.T) {
```
SwampDragons, line 15: I'm not seeing it in here -- isn't it also allowed to access a smb share with just a // instead of a smb://?  should we test these paths?

sylviamoss: This is allowed for windows, I'm not sure if it's allowed for Linux users I think it's not. [Here](https://ubuntu.com/tutorials/install-and-configure-samba#4-setting-up-user-accounts-and-connecting-to-share) there's an Ubuntu tutorial showing how to access smb shared folder from Linux and Windows machines.
Anyways, this case is handled by the FileGetter already, the smb getter is only for when the url has `smb` scheme. 

If the URL is `//dir/file` there's no way to know the files is shared using smb or CIFS protocol, for example, so we assume this is a file path.

I know it works because I tested it locally on Windows, and besides that the smb-getter transforms `smb://host/dir/file` to `//host/dir/file` for windows users and uses the FileGetter to resolve this. There's no test for this because it has to run on Windows. Do you have any thoughts on how to test it?

get_smb.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	// First look in a possible local mount of the shared folder
+	path := "/" + req.u.Host + req.u.Path
+	if runtime.GOOS == "windows" {
+		path = "/" + path
+	}
+	f := new(FileGetter)
+	result := f.getFile(path, req, ctx)
+	if result == nil {
```
sylviamoss, line 178: @SwampDragons This is where the url is transformed from `smb://`  to `//` or `/`, depending on the OS, and used by the FileGetter.

README.md
sylviamoss: I wrote a simple explanation for now but I'll improve after implementing what is missing

README.md
```diff
@@ -360,3 +361,30 @@ In order to access to GCS, authentication credentials should be provided. More i
...
+- smb://username@host/shared/dir
+- smb://host/shared/dir 
+- smb://username:password@host/shared/dir/file (downloads file)
+- smb://username@host/shared/dir/file
+- smb://host/shared/dir/file
+- //host/shared/dir (downloads directory content)
+- //host/shared/dir/file (downloads file)
+
+With and without the smb:// scheme, go-getter will first look for a local mount with file path as `/host/shared/dir` (Unix) or `//host/shared/dir` (Windows), and if the local mount is not available 
+it will then try to download the file/directory via `smbclient` command.
```
sylviamoss, line 381: @azr  do you think is better now?

detect_smb.go
```diff
@@ -0,0 +1,21 @@
...
+
+func (d *SmbDetector) Detect(src, pwd string) (string, bool, error) {
+	if len(src) == 0 {
+		return "", false, nil
+	}
+
+	if strings.HasPrefix(src, "//") {
+		// This is a valid smb path and will be also checked as local file by the SmbGetter
+		return fmt.Sprintf("smb:%s", src), true, nil
+	}
```
sylviamoss, line 19: `//` will be ignored by the FileDetector and will be recognized by the SmbDetector and transformed to a smb scheme URL. This is ok because the SmbGetter uses the FileGetter to check first if there's a possible local mount for the path.

README.md
azr, line 381: I'd may be try to regroup the two explanations together because my fear is that users could just read the first one and be done, I'll to regroup them to my understanding please feel free to fix it :D

```suggestion
The SMB getter will try to get files from samba when a url is prefixed with `smb://` for example `smb://foo/bar/dir`. 

The go-getter will try for a local mount of the path such as `/foo/bar/dir` (Unix) or `//foo/bar/dir` (Windows), and will fallback to using the [`smbclient`](https://www.samba.org/samba/docs/current/man-html/smbclient.1.html) command to download the file.

⚠️ On Unix when the smb share is not mounted, [`smbclient`](https://www.samba.org/samba/docs/current/man-html/smbclient.1.html) needs to be installed for this to work.

Some examples for addressing the scheme:    
- smb://username:password@host/shared/dir (downloads directory content)
- smb://username@host/shared/dir
- smb://host/shared/dir 
- smb://username:password@host/shared/dir/file (downloads file)
- smb://username@host/shared/dir/file
- smb://host/shared/dir/file
- //host/shared/dir (downloads directory content)
- //host/shared/dir/file (downloads file)

 
```

sylviamoss: Love it! Thanks!

mildwonkey:
This looks fine to me as far as terraform core is concerned! Terraform configures a specific list of getters for exactly this reason - no surprises!
(for ex: https://github.com/hashicorp/terraform/blob/807267d1b56aa6298fecf5edba98c088c2866e83/internal/initwd/getter.go#L50)

I'm not necessarily qualified to review the contents of the PR, but I can confirm that this shouldn't impact core.

get_smb.go
```diff
@@ -0,0 +1,340 @@
...
+			dstExisted = true
+		}
+	}
+
+	// First look in a possible local mount of the shared folder
+	path := "/" + req.u.Host + req.u.Path
+	if runtime.GOOS == "windows" {
+		path = "/" + path
+	}
+	f := new(FileGetter)
```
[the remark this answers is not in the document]
SwampDragons: @schmichael as the nomad code currently lives, it appears that this would not be an issue because you're not enabling the SMB getter either, correct?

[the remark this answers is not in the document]
schmichael: Correct! That being said other users of go-getter might be surprised that the `file` getter is implicitly enabled if they enable the `smb` getter.

Perhaps this code path could check if the `file` getter is enabled before deferring to it?

get_smb.go
```diff
@@ -0,0 +1,340 @@
...
+		return false, err
+	}
+	if strings.Contains(output, "OBJECT_NAME_NOT_FOUND") {
+		return false, fmt.Errorf("source path not found: %s", output)
+	}
+	return strings.Contains(output, "attributes: D"), nil
+}
+
+func (g *SmbGetter) runSmbClientCommand(smbclientCmd string, dst string) (string, error) {
+	cmd := exec.Command("bash", "-c", smbclientCmd)
```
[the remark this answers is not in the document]
sylviamoss: Nice! I was struggling to use it as a first argument but then. I was able to do it with your help on Slack. Thanks!

get_smb.go
```diff
@@ -0,0 +1,340 @@
...
+		}
+	}
+
+	// First look in a possible local mount of the shared folder
+	path := "/" + req.u.Host + req.u.Path
+	if runtime.GOOS == "windows" {
+		path = "/" + path
+	}
+	f := new(FileGetter)
+	result := f.getFile(path, req, ctx)
```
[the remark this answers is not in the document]
sylviamoss: Since I'm using `go-multierror` I didn't want to override the previous `err` and result was the multierror variable. I created now an explicitly var to this to make it look better, but with the same name though 🤔

get_smb.go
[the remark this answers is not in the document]
sylviamoss: I've changed a bit based on your suggestions. Now for Windows only, if a FileGetter is available, the SmbGetter will use it to get shared files via the Windows file system.
You're right, It doesn't make sense to allow using the FileGetter for Unix users cuz this can make people get unwanted files. This is the Nomad concern, right? 
For Windows, this still has to be done because as I know the file system is the most common way to access network shared folders.

sylviamoss:
I'm opening another PR with an updated version of this one.

PR #241, opened 2020-04-16
Add windows tests to CircleCI

mdeggies:
This PR adds another job to run the go tests on a Windows executor, which will enable us to get rid of the appveyor dependency. The google key is also stored in a context now so we can remove the project env var once this is merged. To run tests with different versions of go, add the version to the array's on lines 106 and 112. The current version is set to 1.14.1, as the golang choco package for windows only goes up to that version right now. 

The linux and win tests run at the same time, but the Windows job does take a lot longer to finish (~3:30) even with gomod caching enabled. The longest step is installing go, which we need to do because the Windows executor only has go 1.12 installed, while the module requires 1.13+. 

Another note- I'd like to turn these static paths into env variables, but env var injection is not working with Windows executors at the moment.. I'll send in a support ticket to see what's up. 

Screen Shot 2020-04-15 at 8 58 34 PM

Thanks for reviewing!


.circleci/config.yml
```diff
@@ -48,14 +48,66 @@ jobs:
...
-      - go-fmt-and-test
+      - linux-tests:
+          context: go-getter
+          matrix:
+            parameters:
+              go-version: ["1.14.1"]
+          name: linux-test-go-<< matrix.go-version >>
+      - windows-tests:
+          context: go-getter
+          matrix:
```
samsalisbury, line 110: Nice! Not tried this feature yet :D

mdeggies:
Thanks for the pointers! I refactored a bit, so now the 'Install go version' and 'go mod download' steps use the cache now. The time is down to 2 mins (still not amazing, but better): https://circleci.com/gh/hashicorp/go-getter/156. Let me know if it's OK to merge.

.circleci/config.yml
mdeggies: I just found out about it recently! It's great  :)

mdeggies:
Thanks all, I just squashed my commits and will merge!

PR #238, opened 2020-03-14
Create sub modules cmd/go-getter/v2, s3/v2 and gcs/v2

yujunz:

PR #244, opened 2020-04-20
Create sub modules cmd/go-getter/v2, s3/v2 and gcs/v2

### LINKED ISSUE #193: feature request: reduce dependency size

azr:
This PR creates the following submodules for the go-getter:
```
.
├── cmd
│   └── go-getter
├── gcs
└── s3
```

Some testing helpers were used everywhere so I put them in `"helper/testing"`. I also added the `"helper/url".MustParse` function that is used everywhere. Other than that what's unexported is still unexported.

This PR replaces #238 so that we can see circle-ci's results and this uses #241 too to avoid divergences.

close #193


cmd/go-getter/main.go
```diff
@@ -56,6 +58,13 @@ func main() {
 	wg := sync.WaitGroup{}
 	wg.Add(1)
 
+	client := getter.DefaultClient
+
+	getters := getter.Getters
```
mildwonkey, line 63: This looks like a change from the current behavior, where callers can define its own list of getters (`client.Getters`) and only those clients get set (originally in `client_options.go`, when it iterates over the getters if not nil). Am I misreading this change? 

Here's an example from terraform: https://github.com/hashicorp/terraform/blob/807267d1b56aa6298fecf5edba98c088c2866e83/internal/initwd/getter.go#L131

azr: On the `v2` branch the client was split from the request and the getters are set on the client, so here this line from terraform ([link](https://github.com/hashicorp/terraform/blob/807267d1b56aa6298fecf5edba98c088c2866e83/internal/initwd/getter.go#L50-L58)) could become:

```go
var goGetterClient = getter.Client{
	Getters: map[string]getter.Getter{
		"file":  new(getter.FileGetter),
		"gcs":   new(getter.GCSGetter),
		"git":   new(getter.GitGetter),
		"hg":    new(getter.HgGetter),
		"s3":    new(getter.S3Getter),
		"http":  getterHTTPGetter,
		"https": getterHTTPGetter,
	},
}
```

azr: And the call would be : `client.Get(ctx, request)`

mildwonkey: I see it now, thank you! Definitely a misread on my part.

mildwonkey:
This looks fine by me (with my "terraform core" hat on)! Since there's a (tiny) breaking change I would like to see that documented or in the changelog, but that's not blocking this PR. Cheers 🎉

azr:
Thanks for your review. Ah I totally forgot about the changelog; I'll draft a PR right away for the whole `v2` branch, once this is merged.

gcs/get_gcs.go
```diff
@@ -9,16 +9,16 @@ import (
...
 	"cloud.google.com/go/storage"
+	"github.com/hashicorp/go-getter/v2"
 	"google.golang.org/api/iterator"
 )
 
-// GCSGetter is a Getter implementation that will download a module from
+// Getter is a Getter implementation that will download a module from
 // a GCS bucket.
-type GCSGetter struct {
-	getter
```
sylviamoss, line 18: Why did you remove this? 🤔
I'm asking because the SmbGetter is using this to check if there's an available FileGetter. We need to check that In case the user doesn't want to provide the FileGetter and then SmbGetter can't use it.

module_test.go
```diff
@@ -1,37 +1,16 @@
...
-
-	return dir
-}
-
-func tempTestFile(t *testing.T) string {
-	dir := tempDir(t)
-	return filepath.Join(dir, "foo")
-}
-
 func testModule(n string) string {
```
sylviamoss, line 14: Can this go in `helper/testing/utils.go` as well or do you want to add this later to the FileGetter and HttpGetter separately? Since only these two use what's left.

appveyor.yml
```diff
@@ -1,16 +0,0 @@
-version: "build-{branch}-{build}"
```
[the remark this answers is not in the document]
azr: Hehe this change comes from the master branch !

module_test.go
azr: I want the FileGetter and HttpGetter to remain in the main pkg and since they are the only ones using it I am leaving this here.

gcs/get_gcs.go
azr: Ah, good question, I removed it because it is unused and is a bad pattern.

It was initially added in the master branch where the client/request are the same object. It would have allowed to reuse the getters of a request to do things like downloading the checksum from a file. It turns out it was not used.

I think this pattern was pretty bad and was going away from the initial design of this library which was aiming at avoiding a getter from knowing about the client ( so separating the concerns ).

Now if the `SmbGetter` needs this; let's pause a minute to think a little more; I need to re-read your PR with that in mind.

azr:
Okay, merging this one ! Thanks everyone, we are not quite ready to tag this yet so there's probably time to fix/change things around.

PR #245, opened 2020-04-28
Create client GetChecksum

sylviamoss:
This PR makes checksumFromFile private and creates client GetChecksum which will receive a request and from that extract the checksum. 
To allow local files paths (e.g. `./file.sum`), the checksumFromFile method now receives the working dir as a parameter. 

This will close https://github.com/hashicorp/packer/issues/9047 together with https://github.com/hashicorp/packer/pull/9129

### DIFF
--- a/checksum.go
+++ b/checksum.go
@@ -12,7 +12,6 @@ import (
 	"fmt"
 	"hash"
 	"io"
-	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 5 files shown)


azr:
Here's what I think should be public the only public function regarding checksumming:

```go
// GetChecksum extracts the checksum from the `checksum` parameter 
// of the src of the Request
// ex:
//  http://hashicorp.com/terraform?checksum=<checksumValue>
//  http://hashicorp.com/terraform?checksum=<checksumType>:<checksumValue>
//  http://hashicorp.com/terraform?checksum=file:<checksum_url>
// when the checksum is in a file, GetChecksum will first client.Get it
// in a temporary directory, parse the content of the file and finally delete it.
// The content of a checksum file is expected to be BSD style or GNU style.
// For security reasons GetChecksum does not try to get the current working
// directory and relative files will only be found when Request.Pwd is set.
//
// BSD-style checksum:
//  MD5 (file1) = <checksum>
//  MD5 (file2) = <checksum>
//
// GNU-style:
//  <checksum>  file1
//  <checksum> *file2
func (client *Client) GetChecksum(ctx context.Context, req *Request) (*FileChecksum, error)
```

What do you think ?

sylviamoss:
> Also when calling ChecksumFromFile(ctx, req <- I think it is very unclear here what req should be ? A request to get the checksum file ? Or the original get request ? And depending on where we are coming from this could cause confusion and ultimately bugs, I think.

I was concerned about this as well, so I totally agree with you here. 

I like your suggestion for the GetChecksum method. I will give it a try on this one and if I think something else to improve I'll let you know!

checksum.go
```diff
@@ -285,6 +286,58 @@ func (c *Client) ChecksumFromFile(ctx context.Context, checksumURL, checksummedP
...
+// and as a result, relative files will only be found when Request.Pwd is set.
+//
+// BSD-style checksum:
+//  MD5 (file1) = <checksum>
+//  MD5 (file2) = <checksum>
+//
+// GNU-style:
+//  <checksum>  file1
+//  <checksum> *file2
+func (c *Client) GetChecksum(ctx context.Context, req *Request) (*FileChecksum, error) {
```
azr, line 308: Nice !!!
I am just passing by - not sure if this is done and I'm sorry to bother if this was already planned - but I wanted to say make sure to put that code to replace the `extractChecksum` func so we don't keep it as dead code 🙂.

PR #250, opened 2020-05-07
test that arbitrary files cannot be checksummed

schmichael:
New test asserts that only the checksums for retrieved files will be
checked. If a checksum file references other files, those will not be
checked.

This is an important security assertion as checksum files should not
cause go-getter to read arbitrary paths and report if their contents
match a checksum.


schmichael:
@mwhooker Yes! Exactly. Should I expand the test case comment? I tried to keep it brief to match the existing style, but it's definitely a weird test. The goal of the test is the help ensure we don't make a change that could cause a security issue for embedders like Nomad.

mwhooker:
It couldn't hurt. There are a lot of tests in that table, and it could be that the subtlety of this one gets lost in the sea

azr (approved):
LGTM ! Might need a rebase

sonamtenzin2 (approved):

PR #252, opened 2020-05-14
Make checksum func public

azr:
Hello there, we needed that functionality as a standalone one in Packer as we have some internal code that uploads artifacts to a remote instance and we then need to checksum it.

We could have copied it since all the fields of `FileChecksum` are public but I thought this tiny improvement would be a little better.

### DIFF
--- a/checksum.go
+++ b/checksum.go
@@ -48,10 +48,11 @@ func (cerr *ChecksumError) Error() string {
 	)
 }
 
-// checksum is a simple method to compute the checksum of a source file
-// and compare it to the given expected value.
-func (c *FileChecksum) checksum(source string) error {
-	f, err := os.Open(source)
+// Checksum computes the Checksum for file using the hashing algorithm from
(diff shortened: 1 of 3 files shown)


checksum.go
```diff
@@ -67,7 +75,7 @@ func (c *FileChecksum) checksum(source string) error {
 			Hash:     c.Hash,
 			Actual:   actual,
 			Expected: c.Value,
-			File:     source,
+			File:     file,
```
joeandaverde, line 78: This would return a closed file. Is that intended?

joeandaverde: Oh, I misread, file isn't the file it's the path.

veverkap: Would it be better to rename the variable then?

azr: I agree, I think that should probably be `filePath` !

azr:
Thanks everyone ! 🙂

PR #255, opened 2020-05-28
Getters refactoring

sylviamoss:
This is an extension of https://github.com/hashicorp/go-getter/pull/240.

### Client refactoring

The PR refactors the client code to make it try all valid Getters for a given source string. Each Getter will try to detect a valid source string and if the detection is successful it will then try to download the artifact. The client loop over the Getters list until the download is successful.

This doesn't change the behavior of the existing Getters and improves the code. Multiple Getters can be created as good candidates to download the same source string but using different possibilities, whenever this is necessary, such as the SmbClientGetter and the SmbMountGetter, also implemented in this PR. If there's a new and different approach to retrieve an artifact and you still want to fall back to the current Getter. This allows us to add it as a new Getter without affecting the existing one. 

### SMB Getters

This PRs adds a SmbClientGetter and a SmbMountGetter to download files from shared folders using SMB protocol.

This PR also adds the necessary config for running the tests. This includes:

- A new step to the CircleCi that starts two containers, one for the smb server and the other for the go-getter (client), and run the tests inside the go-getter container;
- Dockerfile and docker-compose files to configure both containers and add shared folders with mock files and directories used on go_smb_test.go and client_test.go;
- Makefile to help running the containers and tests locally;
- Readme information about the tests and the SMB protocol.

This is motivated by issues hashicorp/packer#8729 and hashicorp/packer#8552 caused by
hashicorp/packer#7627


client.go
```diff
@@ -75,193 +61,242 @@ func (c *Client) Get(ctx context.Context, req *Request) (*GetResult, error) {
 		req.Dst = td
 	}
 
-	req.u, err = urlhelper.Parse(req.Src)
```
sylviamoss, line 78: All the code from here were moved to inside the Getters for loop

get_smb_mount.go
```diff
@@ -0,0 +1,103 @@
...
+	switch runtime.GOOS {
+	case "windows":
+		prefix = string(os.PathSeparator) + string(os.PathSeparator)
+		path = filepath.Join(u.Host, u.Path)
+	case "darwin":
+		prefix = string(os.PathSeparator)
+		path = filepath.Join("Volumes", u.Path)
+	case "linux":
+		prefix = string(os.PathSeparator)
+		path = fmt.Sprintf("run/user/1000/gvfs/smb-share:server=%s,share=%s", u.Host, strings.TrimPrefix(u.Path, prefix))
```
sylviamoss, line 65: Not sure about this one. I'm thinking maybe Linux user should always have smbclient installed for this to work.

get_smb_mount.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (g *SmbMountGetter) findPrefixAndPath(u *url.URL) (string, string) {
+	var prefix, path string
+	switch runtime.GOOS {
+	case "windows":
+		prefix = string(os.PathSeparator) + string(os.PathSeparator)
+		path = filepath.Join(u.Host, u.Path)
+	case "darwin":
+		prefix = string(os.PathSeparator)
+		path = filepath.Join("Volumes", u.Path)
```
sylviamoss, line 62: These don't have smbclient command available

cmd/go-getter/main.go
```diff
@@ -61,8 +61,8 @@ func main() {
 	client := getter.DefaultClient
 
 	getters := getter.Getters
-	getters["gcs"] = new(gcs.Getter)
-	getters["s3"] = new(s3.Getter)
+	getters = append(getters, new(gcs.Getter))
+	getters = append(getters, new(s3.Getter))
```
azr, line 65: Do we want to add the smb getter as as default getter cmd here ? ( I would think so )

sylviamoss: They are already in the `getter.Getters` as the rest of the Getters

get.go
```diff
@@ -66,12 +72,13 @@ func init() {
...
-		"file":  new(FileGetter),
-		"git":   new(GitGetter),
-		"hg":    new(HgGetter),
-		"http":  httpGetter,
-		"https": httpGetter,
+	Getters = []Getter{
+		new(GitGetter),
+		new(HgGetter),
+		new(SmbClientGetter),
+		new(SmbMountGetter),
```
sylviamoss, line 79: Here they are @azr

cmd/go-getter/main.go
azr: Right !

client.go
```diff
@@ -75,193 +61,242 @@ func (c *Client) Get(ctx context.Context, req *Request) (*GetResult, error) {
...
+
+			// Process any globs
+			subDir, err := SubdirGlob(req.Dst, subDir)
+			if err != nil {
+				return nil, err
+			}
+
+			return &GetResult{realDst}, copyDir(ctx, realDst, subDir, false)
 		}
+		return &GetResult{req.Dst}, nil
```
azr, line 259: I think it would be better for readability and for this diff if the content of that loop was put into a func:

May be something like:

```go
for _, g := range c.Getters {
	op, err, := c.get(...)...
	// ...
}
```

azr: Also it will be easier to see what happens at the error management level etc.

sylviamoss: Good idea!!

detect_bitbucket.go
```diff
@@ -1,66 +0,0 @@
...
-	"encoding/json"
-	"fmt"
-	"net/http"
-	"net/url"
-	"strings"
-)
-
-// BitBucketDetector implements Detector to detect BitBucket URLs and turn
-// them into URLs that the Git or Hg Getter can understand.
-type BitBucketDetector struct{}
```
azr, line 13: Here I was wondering if we should may be make the git getter have the ability to have multiple detectors like this one. For example:

```go
GitGetter{
	Detectors: []{BitBucketDetector, GithubDetector, ...}
}
```

This would largely reduce the diff of this PR and sort of reduce friction for anyone reviewing this and for git history. Because the current detectors will be sort of kept very similar and in the same place.

🤔 What do you think

detect_test.go
```diff
@@ -11,81 +11,82 @@ func TestDetect(t *testing.T) {
...
 	}
 
 	for i, tc := range cases {
 		t.Run(fmt.Sprintf("%d %s", i, tc.Input), func(t *testing.T) {
-			output, err := Detect(tc.Input, tc.Pwd, Detectors)
+			req := &Request{
+				Src: tc.Input,
+				Pwd: tc.Pwd,
+			}
+			_, err := Detect(req, tc.getter)
```
azr, line 84: Should we check for that boolean return here too ?

get.go
azr: Yes I was looking for that, thx !

azr: Ah btw I think we should add a comment here stating that the order is importer

client.go
```diff
@@ -19,18 +20,13 @@ import (
...
-	Detectors []Detector
-
 	// Decompressors is the map of decompressors supported by this client.
 	// If this is nil, then the default value is the Decompressors global.
 	Decompressors map[string]Decompressor
 
 	// Getters is the map of protocols supported by this client. If this
 	// is nil, then the default Getters variable will be used.
-	Getters map[string]Getter
+	Getters []Getter
```
azr, line 29: This comment needs to be updated here too

get.go
```diff
@@ -42,13 +42,20 @@ type Getter interface {
 	// Mode returns the mode based on the given URL. This is used to
 	// allow clients to let the getters decide which mode to use.
 	Mode(context.Context, *url.URL) (Mode, error)
+
+	// Detect detects whether the Request.Src matches a known pattern to
+	// turn it into a proper URL.
+	// The Getter must validate if the Request.Src is a valid URL
+	// with a valid scheme for the Getter, and also check if the
+	// current Getter is the forced one and return true if that's the case.
```
azr, line 50: Should we say here that some getters actually can change the Request ?

get_file_unix_test.go
```diff
@@ -55,16 +55,20 @@ func TestFileDetector_relativeSymlink(t *testing.T) {
...
 	// if detech doesn't fully resolve the pwd symlink, the output will be the
 	// invalid path: "file:///../modules/foo"
-	f := new(FileDetector)
-	out, ok, err := f.Detect("../modules/foo", "./linkedPWD")
+	f := new(FileGetter)
+	req := &Request{
+		Src: "../modules/foo",
+		Pwd: "./linkedPWD",
+	}
+	ok, err := f.Detect(req)
```
azr, line 63: Nice ! 👍

get_smb_mount.go
azr: Should this be something like `run/user/$uid/gvfs` in case the user is not 1000 ?

get_smbclient_test.go
```diff
@@ -0,0 +1,327 @@
...
+		})
+	}
+}
+
+func smbTestsPreCheck(t *testing.T) {
+	r := os.Getenv("ACC_SMB_TEST")
+	if r != "1" {
+		t.Skip("smb getter tests won't run. ACC_SMB_TEST not set")
+	}
+}
```
azr, line 327: what do you think about putting this func at the top of the file with a comment so that users understands that all tests in this file can be skipped here.

Also, super good job on the smb server in circle.

client.go
```diff
@@ -50,43 +46,61 @@ func (c *Client) Get(ctx context.Context, req *Request) (*GetResult, error) {
...
 
-	g, ok := c.Getters[force]
-	if !ok {
-		return nil, fmt.Errorf(
-			"download not supported for scheme '%s'", force)
+	return nil, fmt.Errorf("error downloading '%s'", req.Src)
+}
+
+func (c *Client) get(ctx context.Context, req *Request, g Getter) (*GetResult, bool, error) {
+	u, err := urlhelper.Parse(req.Src)
```
sylviamoss, line 100: here the code remains the same as before

detect_git_test.go
```diff
@@ -41,28 +41,22 @@ func TestGitDetector(t *testing.T) {
 			"git@github.xyz.com:org/project.git//module/a?ref=test-branch",
 			"git::ssh://git@github.xyz.com/org/project.git//module/a?ref=test-branch",
 		},
-		{
-			// Already in the canonical form, so no rewriting required
-			// When the ssh: protocol is used explicitly, we recognize it as
-			// URL form rather than SCP-like form, so the part after the colon
-			// is a port number, not part of the path.
-			"git::ssh://git@git.example.com:2222/hashicorp/foo.git",
```
sylviamoss, line 49: This test was moved to get_git_test.go

client.go
```diff
@@ -50,43 +46,61 @@ func (c *Client) Get(ctx context.Context, req *Request) (*GetResult, error) {
...
+	if multierr != nil {
+		var result *multierror.Error
+		result = multierror.Append(result, multierr...)
+		return nil, fmt.Errorf("error downloading '%s': %s", req.Src, result.Error())
+	}
+
+	return nil, fmt.Errorf("error downloading '%s'", req.Src)
+}
+
+func (c *Client) get(ctx context.Context, req *Request, g Getter) (*GetResult, bool, error) {
```
sylviamoss, line 99: This returns a bool to tell if the error is fatal or not. 
This is because some errors are fatal, I mean when something is wrong with the source string, for these we return `nil, true, err`. Besides these errors, is also possible that the actual download will fail inside the getter and when this happens we want to try any other possible getter in the getters list, so we return `nil, false, err` and save the error to return later if the download is not successful at all.

client.go
```diff
@@ -50,43 +46,61 @@ func (c *Client) Get(ctx context.Context, req *Request) (*GetResult, error) {
...
+		req.realDst = req.Dst
 		req.Dst = td
 	}
 
-	req.u, err = urlhelper.Parse(req.Src)
-	if err != nil {
-		return nil, err
+	var multierr []error
+	for _, g := range c.Getters {
+		ok, err := Detect(req, g)
```
azr, line 65: Here it is unclear what `ok` refers to, may be something like `shouldDownload` or `matches` could be better in this context ?

client.go
```diff
@@ -50,43 +46,61 @@ func (c *Client) Get(ctx context.Context, req *Request) (*GetResult, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		if !ok {
+			continue
```
azr, line 70: What do you think about: 
```suggestion
		if !ok {
			// the request should not be processed by that getter
			continue
```

s3/get_s3.go
```diff
@@ -272,3 +271,48 @@ func (g *Getter) parseUrl(u *url.URL) (region, bucket, path, version string, cre
...
+	src, ok, err := new(Detector).Detect(src, req.Pwd)
+	if err != nil {
+		return ok, err
+	}
+	if ok {
+		req.Src = src
+		return ok, nil
+	}
+
+	return false, nil
```
azr, line 313:
```suggestion
	if ok {
		req.Src = src
	}

	return ok, nil
```

client.go
```diff
@@ -19,18 +20,13 @@ import (
...
-
 	// Decompressors is the map of decompressors supported by this client.
 	// If this is nil, then the default value is the Decompressors global.
 	Decompressors map[string]Decompressor
 
-	// Getters is the map of protocols supported by this client. If this
+	// Getters is the list of protocols supported by this client. If this
 	// is nil, then the default Getters variable will be used.
-	Getters map[string]Getter
+	Getters []Getter
```
azr, line 29: nice

client.go
azr: It feels like this bool is here to tell when the calling loop needs to stop.

May `get` could return a specific error:

```go
type getError struct {
	// when fatal is true it means that something when wrong with
	// the get and that the getter knows nothing more should be tried
	Fatal	bool
	Error	error
}

func (ge *GetError) Error() string {
	// ...
}
```

```suggestion
func (c *Client) get(ctx context.Context, req *Request, g Getter) (*GetResult, *getError) {
```

detect_file.go
```diff
@@ -56,12 +56,6 @@ func fmtFileURL(path string) string {
...
 		// Make sure we're using "/" on Windows. URLs are "/"-based.
 		path = filepath.ToSlash(path)
-		return fmt.Sprintf("file://%s", path)
 	}
-
-	// Make sure that we don't start with "/" since we add that below.
-	if path[0] == '/' {
-		path = path[1:]
-	}
-	return fmt.Sprintf("file:///%s", path)
```
azr, line 66: Was this causing a bug ? What makes this change possible ?

request.go
```diff
@@ -18,6 +18,10 @@ type Request struct {
 	Dst string
 	Pwd string
 
+	// Forced getter detected in Src during Getter detection phase
+	// This is currently for internal use only
+	Forced string
+
```
azr, line 24: Ah, I wish things around that `Forced` field where a little less funky. 

Can you please document a little more things around this field ? I keep forgetting why we need that field 😓 .

gcs/detect_gcs_test.go
```diff
@@ -11,20 +11,20 @@ func TestGCSDetector(t *testing.T) {
...
 		},
 		{
 			"www.googleapis.com/storage/v1/bucket/foo/bar",
-			"gcs::https://www.googleapis.com/storage/v1/bucket/foo/bar",
+			"https://www.googleapis.com/storage/v1/bucket/foo/bar",
 		},
 		{
 			"www.googleapis.com/storage/v1/foo/bar.baz",
-			"gcs::https://www.googleapis.com/storage/v1/foo/bar.baz",
+			"https://www.googleapis.com/storage/v1/foo/bar.baz",
```
azr, line 22: Will a `gcs::` get still work ?

get_smb_mount.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+	return prefix, path
+}
+
+func (g *SmbMountGetter) Detect(req *Request) (bool, error) {
+	if runtime.GOOS == "linux" {
+		// Linux users should use smbclient command.
```
azr, line 66: Can you please explain why in the comment ( In case the why gets fixed in the future )

get_smb_mount.go
```diff
@@ -0,0 +1,101 @@
...
+
+import (
+	"context"
+	"net/url"
+	"os"
+	"path/filepath"
+	"runtime"
+)
+
+// SmbMountGetter is a Getter implementation that will download a module from
```
azr, line 11:
```suggestion
// SmbMountGetter is a Getter implementation that will download an artifact from
```

detect_file.go
sylviamoss: This was only for telling the client that the FileDetector detected the source string as valid. In the previous code, once this is back to the client, the client was going to remove the forced "file" to use as a key to look for the getter on the map. No longer necessary.

gcs/detect_gcs_test.go
sylviamoss: Yes. This is the same situation as the FileDetector you asked above. This was for the client.go to know which getter to use.

detect_file.go
azr: Oh right ! 💯  thanks for explaining !

azr:
Nice, this is such a big PR 🙂  ! I tried looking at everything with a bit more attention to the smb code and nice !

get_smb_mount.go
sylviamoss: Do you mean why to use smbclient command? Like for example, because it is safer?

client.go
```diff
@@ -39,6 +35,17 @@ type GetResult struct {
 	Dst string
 }
 
+type getError struct {
+	// when fatal is true it means that something when wrong with
```
nywilken, line 39: It would be good to have this comment added to any exported method that returns a getError so that it is clear what to expect from the error. Or maybe just export this type as in `GetError` and add the comment above the type so that it is available via godoc. What do you think?

sylviamoss: Right now this one is only returned by the `get` method which is not exported. Do you think exporting `getError` is necessary anyways?
I'm updating the comment to make it clear that is used by `get`.

nywilken: Oh in that case you can disregard my comment. I don't think exporting it is necessary if its only used by unexported methods.

PR #256, opened 2020-05-28
umask final version against master (again)

shoenig:
This is https://github.com/hashicorp/go-getter/pull/198 but rebased with master as of May 2020.

Nomad needs a recent fix for https://github.com/hashicorp/nomad/issues/7854

cc @langmartin


notnoop (approved):
LGTM and this has been running as-in in nomad for many versions now.

azr:
Nice one !

PR #259, opened 2020-06-07
feat(detector): provide a detector for repository hosted on GitLab.com

davinkevin:
This is a really simple implementation based on the GitHub detector for GitLab URL. It only supports URL with format `gitlab.com/username/repo`

### DIFF
--- a/detect_gitlab.go
+++ b/detect_gitlab.go
@@ -1,0 +1,47 @@
+package getter
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// GitLabDetector implements Detector to detect GitLab URLs and turn
+// them into URLs that the Git Getter can understand.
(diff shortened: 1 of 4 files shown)


detect_gitlab.go
```diff
@@ -0,0 +1,47 @@
+package getter
+
+import (
+	"fmt"
+	"net/url"
+	"strings"
+)
+
+// GitHubDetector implements Detector to detect GitHub URLs and turn
```
sysadmiral, line 9: Please update comment to replace GitHub with Gitlab

azr, line 9:
```suggestion
// GitLabDetector implements Detector to detect GitLAb URLs and turn
```

davinkevin: Done

azr:
Thanks !

PR #266, opened 2020-08-28
Let shallow clone work with any ref

zachwhaley:
Currently the shallow clone only works if ref is either not set, or set
to the default branch.  This is becuase go-getter does not set the ref
to checkout during a shallow clone so Git assumes the default branch,
and then when go-getter tries to checkout a ref other than the default
branch it fails because the shallow clone did not fetch that ref.

To fix this, I have used `--branch` to explicitly set the ref to fetch
during a clone so that when a shallow clone is done go-getter will clone
the ref given and not the default branch.  This will also make clones of
non-default branches just a tad bit more optimized since the clone will
not waste time checking out a branch it ultimately will not use.

Fixes https://github.com/hashicorp/terraform/issues/10703


zachwhaley:
I was not able to get the tests to run locally and I couldn't seem to find any documentation to explain this.  Also, I'm not sure if I was supposed to open a PR here or in terraform.

Please advise :)

zachwhaley:
Hello?

zachwhaley:
I see that a similar PR was opened 9 months ago to address the same thing #251

Is this repo abandoned?

zachwhaley:
Hello?

sebglon:
@schmichael can we have your helm on this PR?
This is a big locket on many project

simoncaron:
@azr Any chance you could take a few minutes to look at this? It would be really great if we could close this issue which has been ongoing for more than a year. Thanks!

schmichael:
Sorry for the delay everyone and thanks for the contribution. Unfortunately you'll still have to wait for dependent projects (Nomad, Terraform, etc) to update their go-getter dependency.

tomqwpl:
On the face of it, the fix for this has changed the behaviour with regard to the "ref" parameter when getting from git. Previously it allowed any ref, so commit hash etc. Now it only allows branch names or tag names. My guess is that this is an unintended side effect of the change?
Doing "git clone; git checkout <ref>" (which is what happened before) isn't the same as doing "git clone --branch <ref>; git checkout <ref>" (which is what happens now), since the "git clone --branch <ref>" bit only works for branch names or tag names.

PR #269, opened 2020-08-28
Add support for 'git::' force token on local filepaths, both absolute and relative

### LINKED ISSUE #268: add support for 'git::' force support on local filepaths, both absolute and relative
It would be cool if the `git::` forcing token could be used on local file system paths, both absolute paths and relative paths. For example:
```
    git::./some/relative/path/to/a/git-repo//some-subdir?ref=v1.2.3
```
or:
```
    git::/some/absolute/path/to/a/git-repo//some-subdir?ref=v4.5.6
```

Internally, `go-getter` should transform the provided string into a `file://` URI with an absolute filepath, with query string params and subdirectory notation retained.

The rationale for using a `file://` URI internally is that the Git clone operation can work with `file://` URIs, and using them for this feature would allow us to leverage the existing `go-getter` URI-handling machinery. That would get us support for query params (to clone a specific git ref (tag, commit hash, ...)) "for free".

The rationale for using an absolute filepath (even when the provided string is a relative filepath) is that (per [RFC 1738](https://tools.ietf.org/html/rfc1738) and [RFC 8089](https://tools.ietf.org/html/rfc8089)) only absolute filepaths are legitimate in `file://` URIs. But more importantly, the Git clone operation only supports `file://` URIs with absolute paths.

* **Q: Why support this functionality at all?**

  Why not just require that a source location use an absolute path in a `file://` URI explicitly if that's what is needed?

* **A:** The primary reason is to allow support for relative filepaths to Git repos.

  There are use cases in which the absolute path cannot be known in advance, but a relative path to a Git repo is known.

  For example, when a Terraform project (or any Git-based project) uses [Git submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules), it will know the relative location of the Git submodule repos, but cannot know the absolute path in advance because it will vary based on where the "superproject" repo is cloned. Nevertheless, those relative paths should be usable as clonable Git repos, and this mechanism would allow for that.

  Support for filepaths that are already absolute would be provided mainly for symmetry. It would be surprising for the feature to work with relative file paths, but not for absolute filepaths.

For projects using [Terraform](https://github.com/hashicorp/terraform), in particular, this feature would enable the non-fragile use of relative paths in a `module` "call" block, when combined with Git submodules:

```terraform
    module "aljunk_parent" {
        source = "git::../git-submodules/tf-modules/some-tf-module?ref=v0.1.0"
        // ...
    }
```

In the above example "superproject" Git repo (the one "calling" the [terraform module](https://www.terraform.io/docs/configuration/modules.html)) knows the relative path to its own Git submodules because they are embedded in a subdirectory beneath the top-level of the "superproject" repo.

There are at least two downstream Terraform issues open that would require `go-getter` support for this feature:

   * hashicorp/terraform#25488: "Unable to use relative path to local Git module"
   * hashicorp/terraform#21107: "In 0.12, modules can no longer be installed from local git repositories at relative paths"

I have a branch that implements this feature. I'll open a PR to get feedback on it shortly.

salewski:
This PR is a WIP, but please feel free to provide feedback.

Fixes #268 

Remaining work:
- [ ] document feature (in README.md)
- [ ] squash down to a single commit
- [ ] document limitations and/or possible future work


salewski:
In the wake of the discussion from earlier today, I've improved my testing setup for this issue and have one that demonstrates the main brokenness of the `ads/git-forced-rel-path-to-abs` branch as it currently exists (at [commit: 31e197d84f27](https://github.com/hashicorp/go-getter/pull/269/commits/31e197d84f27ce266306ebc7dcf1b0702d5465a2)):

Here's the repo structure in which I've produced the problem:
```       
  1    $ tree ads-exp-tf-parent/
  2    ads-exp-tf-parent/                <= 'terraform init' invoked here
  3    ├── README.md
  4    ├── git-submodules
  5    │   └── tf-modules
  6    │       └── ads-exp-tf-module     <= Git repo (git submodule)
  7    │           ├── README.md
  8    │           ├── main.tf
  9    │           ├── outputs.tf
 10    │           └── variables.tf
 11    ├── local-tf-modules
 12    │   └── foo
 13    │       ├── local-tf-modules
 14    │       │   └── bar
 15    │       │       ├── main.tf       <= FAILS: "git::../../../../git-submodules/tf-modules/ads-exp-tf-module?ref=v0.1.1"
 16    │       │       ├── outputs.tf
 17    │       │       └── variables.tf
 18    │       ├── main.tf
 19    │       ├── outputs.tf
 20    │       └── variables.tf
 21    ├── main.tf                       <= WORKS: "git::./git-submodules/tf-modules/ads-exp-tf-module?ref=v0.1.0"
 22    ├── outputs.tf
 23    ├── terraform.tfstate
 24    ├── terraform.tfstate.backup
 25    └── variables.tf
```

The directory `ads-exp-tf-parent` is a Git repository. It is a "superproject" that contains Git submodules (see [git-submodule(1)](https://git-scm.com/docs/git-submodule)).
       
The Git repo reference on line 15 fails because the specified relative path is correct relative to the location of the `bar/main.tf` file, but incorrect relative the `$PWD` of the `terraform` process.
       
The Git repo reference on line 21 works because the specified path happens to be correct relative to the `$PWD` of the `terraform` process.

For reference, the specific error emitted by (my locally modified) `terraform` is:
```
    e:
    Error: Failed to download module

    Could not download module "bar_breaker"
    (local-tf-modules/foo/local-tf-modules/bar/main.tf:7) source code from
    "git::../../../../git-submodules/tf-modules/ads-exp-tf-module?ref=v0.1.1":
    error downloading
    'file:///git-submodules/tf-modules/ads-exp-tf-module?ref=v0.1.1': /usr/bin/git
    exited with 128: Cloning into
    '.terraform/modules/aljunk_sub1.aljunk_sub2.bar_breaker'...
    fatal: '/git-submodules/tf-modules/ads-exp-tf-module' does not appear to be a
    git repository
    fatal: Could not read from remote repository.

    Please make sure you have the correct access rights
    and the repository exists.
```

lae:
Any chance we could get a review of this soon?

lorengordon:
I love this idea!

aellwein:
any progress here??

PR #275, opened 2020-09-24
Add role option for S3 URLs

02strich:
Add support for specifying an IAM role that is assumed before making S3
requests. Beforehand this was only possible when using profiles which
required local files for configuration.

Now it is possible to specify such an IAM role in the URL, thereby
supporting settings where a IAM role needs to be assumed for data
access (e.g. cross-account).

Example of usage:
    s3::bucket.s3.amazonaws.com/file?aws_iam_role=some-role-arn

This currently only has a negative test case as I do not have access to the used AWS account. If someone could help me with details to write a positive test case as well, that would be much appreciated.


02strich:
Any chance to get some feedback on this?

02strich:
Any chance to get someone to look at this?

02strich:
ping?

PR #277, opened 2020-09-28
Add support for github releases

mdeggies:
This PR adds support for github releases on tag pushes. It includes code signing for darwin/win packages, GPG signing for shasum validation, and malware/AV scanning for good measure. 

close #115


azr (approved):
Super cool ! LGTM !

azr:
This is going to close #115 so I added that in the PR description

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+          sha256sum -c - <<< "$SHA256SUM  maldet-$VERSION.tar.gz"
+          sudo mkdir -p maldet-$VERSION
+          sudo tar -xzf maldet-$VERSION.tar.gz --strip-components=1 -C maldet-$VERSION
+          cd maldet-$VERSION
+          sudo ./install.sh
+          sudo maldet -u
+        env: 
+          VERSION: 1.6.4
+          SHA256SUM: 3ad66eebd443d32dd6c811dcf2d264b78678c75ed1d40c15434180d4453e60d2
+      - name: Import PGP key for archive signing
```
samsalisbury, line 45: Since this is OSS, are we sure we want to be doing this here? I'm not very familiar with GH Actions security model, so this might be fine.

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+          docker pull docker.pkg.github.com/hashicorp/hc-codesign/hc-codesign:$VERSION && \
+          echo "::set-output name=image::docker.pkg.github.com/hashicorp/hc-codesign/hc-codesign:$VERSION"
+        env:
+          VERSION: v0
+          GITHUB_TOKEN: ${{ secrets.CODESIGN_GITHUB_TOKEN }}
+      - name: Install wget & clamAV antivirus scanner
+        run : |
+           sudo apt-get -qq install -y ca-certificates wget clamav
+           wget --version
+      - name: Install maldet malware scanner
```
samsalisbury, line 33: I would recommend running the full install script, because it includes [a self-test that checks both scanners are actually working](https://github.com/hashicorp/packagespec/blob/master/assets/private/release/scripts/av-install#L68-L95). I think this is important because the scanner pulls down fresh definitions when it is run, so its behaviour is not guaranteed to be consistent. For packagespec I copied this whole script https://github.com/samsalisbury/av-scan/blob/master/install to `assets/private/release/scripts/av-install` and just call that from CI.

.goreleaser.yml
```diff
@@ -0,0 +1,49 @@
+before:
+  hooks:
+    - go test ./...
+
+builds:
+  - id: signable
+    mod_timestamp: '{{ .CommitTimestamp }}'
```
samsalisbury, line 7: I might steal this idea for packagespec to go towards reproducible builds :)

.goreleaser.yml
```diff
@@ -0,0 +1,49 @@
...
+    targets:
+      - darwin_amd64
+      - windows_386
```
samsalisbury, line 10: Are we signing the Windows 386 builds? I didn't think we were, and so packagespec does not at the moment, but if that's wrong I need to change it ASAP!

.github/workflows/release.yml
```diff
@@ -0,0 +1,70 @@
...
+          fetch-depth: 0
+      - name: Setup go
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.15'
+      - name: Install hc-codesign
+        id: codesign
+        run: |
+          docker login docker.pkg.github.com -u docker -p $GITHUB_TOKEN && \
+          docker pull docker.pkg.github.com/hashicorp/hc-codesign/hc-codesign:$VERSION && \
```
samsalisbury, line 24: I did not know about this image! Is this something we can use elsewhere as well?

.goreleaser.yml
```diff
@@ -0,0 +1,49 @@
...
+    flags:
+      - -trimpath
+    ldflags:
+      - '-X "main.GitCommit=$(shell git rev-parse HEAD)"'
+
+archives:
+  - format: zip
+    name_template: "{{ .ProjectName }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}"
+    files: 
+      - none*
```
samsalisbury, line 38: What does this `- none*` mean?

samsalisbury (commented):
Hi @mdeggies, no blockers here I think, but I do have a few questions/comments inline. Also, there's some stuff I want to steal from this PR :)

samsalisbury (commented):
Hi @mdeggies, no blockers here I think, but I do have a few questions/comments inline. Also, there's some stuff I want to steal from this PR :)

.github/workflows/release.yml
mdeggies: Yeah, this is fine, as all secrets are obfuscated. Many of our other OSS projects are doing this too!

.goreleaser.yml
mdeggies: Yes, we are signing win 386 builds (as well as amd64 + darwin amd64). I'll go through the other products this week to make sure they're doing this as well

.github/workflows/release.yml
mdeggies: Yes! The repo has instructions. Thanks to radek for creating this :)

.goreleaser.yml
mdeggies: I'll add a comment here- it just removes any extraneous/default files before packaging

.github/workflows/release.yml
mdeggies: Thanks Sam, I wasn't sure why these checks were here. If maldet or clamav aren't functioning properly, will the calls to them error out, or are there other 'hidden' failures that this will capture?

samsalisbury: The idea of the checks is to see whether we can trust the exit code from both tools. We first feed it a test blob that contains a known string that should be identified as malware and ensure that the exit code is not zero. Then feed it a benign blob and check that the exit code is zero.

samsalisbury (approved):

PR #261, opened 2020-06-18
Support query parameter aws_profile for S3

graywolf-at-work:
Add support for aws_profile query parameter to allow fetching from S3
using that specified profile. Before this, that did not seem possible,
even setting AWS_PROFILE environment variable did nothing.

Now it is possible to provide the profile directly in the url, making it
more flexible.

Example of usage:

	s3::bucket.s3.amazonaws.com/file?aws_profile=some-profile

PR #282, opened 2020-09-30
Update v2 branch with code from the master branch

azr:
this brings #256 & #261 into the v2 branch

I had rename Request.Mode to Request.GetMode so that we can have a Request.Mode ( for the umask mode ). Maybe there's a better name than Request.Mode there too


s3/get_s3.go
```diff
@@ -310,3 +307,25 @@ func (g *Getter) Detect(req *getter.Request) (bool, error) {
...
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		config := g.getAWSConfig(region, url, creds)
+		sess = session.New(config)
+	}
+
+	return s3.New(sess), nil
+}
```
azr, line 331: this is from #261

PR #284, opened 2020-10-20
Fix bug in Detect() where a Windows-style absolute path is not translated properly

brianpursley:
When calling `Detect()` with a Windows-style absolute path, it is detected to be a valid URL by `url.Parse()` and returned directly, without applying the `file://` prefix, like other file paths.

This PR excludes file scheme URLs from directly returning if successfully parsed, allowing them to flow down and be handled by `FileDetector`, which takes care of applying the `file://` prefix.

This PR adds unit test cases covering Windows-style path scenarios.


brianpursley:
@mdeggies can you take a look at this PR when you get a chance?

PR #285, opened 2020-10-20
parse checksum even when no EOL is found

nixpanic:
In case a checksum file does not end with '\n', the checksum can still
be valid. When hitting EOF, verify that the read line is non-empty, and
continue with parsing.

This addresses kubernetes/minikube#9470.


azr (commented):
Hello there, nice one, this LGTM ! Could you please add a test case for this one as well ?

nixpanic:
Testcase added, without the change it FAILS:
```
$ go test . -test.run TestGetFile_checksum_from_file
--- FAIL: TestGetFile_checksum_from_file (0.02s)
    --- FAIL: TestGetFile_checksum_from_file/?checksum=file:http://127.0.0.1:44575/sha512-p-EOF.sum (0.00s)
        get_test.go:444: append: ?checksum=file:http://127.0.0.1:44575/sha512-p-EOF.sum
            
            err: invalid checksum: no checksum found in: http://127.0.0.1:44575/sha512-p-EOF.sum
FAIL
FAIL	github.com/hashicorp/go-getter	0.027s
FAIL
```

And with the change it passes:
```
$ go test . -test.run TestGetFile_checksum_from_file
ok  	github.com/hashicorp/go-getter	0.019s
```

azr (approved):

azr:
Good work, thanks !

afbjorklund:
Sad that this is needed, but thanks for the workaround! The kubernetes release still might not verify using other tools though

PR #296, opened 2020-12-23
enable code scanning for v2 branch

chair6:
Adopting the same code scanning config for `v2` as was brought into the `main` branch with https://github.com/hashicorp/go-getter/pull/295.


chair6:
Will do this with a cherry-pick instead.

azr:
Thanks !

PR #301, opened 2021-01-07
Set GOPRIVATE env var

mdeggies:
The goreleaser release pipeline that was created and used for the 1.5.1 release was missing the GOPRIVATE env var being set, which caused private modules to be skipped. This change is being made [in code](https://goreleaser.com/customization/env/) so that it's visible to users. After it's merged, we'll kick off the v1.5.2 release.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,3 +1,6 @@
+env:
+  - GOPRIVATE=github.com/hashicorp
+
 builds:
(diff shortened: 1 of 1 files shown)


99:
can't review the context, but change is lg

99 (approved):

azr (approved):

PR #310, opened 2021-02-25
Add support for gcs object generations

ScottSuarez:
Adding support for gcs generations. These are essentially versioned objects in gcs buckets annotated by [a hash at the end of the url](https://cloud.google.com/storage/docs/object-versioning).

### DIFF
--- a/get_gcs.go
+++ b/get_gcs.go
@@ -6,7 +6,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 
@@ -135,8 +134,7 @@ func (g *GCSGetter) GetFile(dst string, u *url.URL) error {
 func (g *GCSGetter) getObject(ctx context.Context, client *storage.Client, dst, bucket, object, fragment string) error {
 	var rc *storage.Reader
 	var err error
-	fragmentHasGeneration := regexp.MustCompile("^\\d+$").MatchString(fragment)
-	if fragmentHasGeneration {
+	if fragment != "" {
 		generation, err := strconv.ParseInt(fragment, 10, 64)
 		if err != nil {
 			return err
--- a/get_gcs.go
+++ b/get_gcs.go
@@ -124,7 +124,6 @@ func (g *GCSGetter) GetFile(dst string, u *url.URL) error {
 	if err != nil {
 		return err
 	}
-	fmt.Println(object)
 
 	client, err := storage.NewClient(ctx)
 	if err != nil {
--- a/get_gcs.go
+++ b/get_gcs.go
@@ -6,6 +6,8 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"cloud.google.com/go/storage"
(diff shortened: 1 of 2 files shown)


get_gcs.go
```diff
@@ -127,11 +129,22 @@ func (g *GCSGetter) GetFile(dst string, u *url.URL) error {
...
-	return g.getObject(ctx, client, dst, bucket, object)
+	return g.getObject(ctx, client, dst, bucket, object, fragment)
 }
 
-func (g *GCSGetter) getObject(ctx context.Context, client *storage.Client, dst, bucket, object string) error {
-	rc, err := client.Bucket(bucket).Object(object).NewReader(ctx)
+func (g *GCSGetter) getObject(ctx context.Context, client *storage.Client, dst, bucket, object, fragment string) error {
+	var rc *storage.Reader
+	var err error
+	fragmentHasGeneration := regexp.MustCompile("^\\d+$").MatchString(fragment)
```
SwampDragons, line 138: Is the fragment able to ever have anything _other_ than generation/version information?  (I know nothing about GCS objects). If so, is this information ever useful?

SwampDragons (approved):
So with the caveat that I know nothing about GCS, the implementation here looks fine to me. 

My only question is whether the URL fragment, which previously was totally ignored, could contain other information than the versioning information, and whether that information being present and suddenly used to try to apply a version could cause a new failure.  I expect the answer is "no" (and reading the docs you linked this seems to be the case) but it's the only potential failure case that pops to mind with this code.

get_gcs.go
schmichael: I'll take the `if not` case :)

If the fragment can only ever be a generation number, can we skip the regex and go straight to ParseInt?

```go
if len(fragment) > 0 {
  // has generation code here
} else {
  // no generation code here
}
```

My concern is that people will intend to get a specific generation with a fragment like: `#v36` but instead of an error due to the `v`, they will silently be given _another_ generation than they expected.

schmichael (changes requested):
Needs a test in `get_gcs_test.go`. Instructions here: https://github.com/hashicorp/go-getter#gcs-testing

get_gcs.go
SwampDragons: Oh great point. Seconded.

ScottSuarez: not that I know of !

ScottSuarez: Well I suppose the reason I have this check here is if the user puts junk in the hash it will prevent the parsing the of the hash as an int64 from erroring out. It's sort of a quality of life thing. If the hash is junk, just pull gcs object remotely without considering the generation. Thats why I decided to include the check. What do you think?

ScottSuarez: Thats a good point about that giving an error if the generation is given in another format ! I didn't think about that. I agree with your perspective!

ScottSuarez:
Added test and revised code

get_gcs_test.go
```diff
@@ -88,6 +89,47 @@ func TestGCSGetter_GetFile(t *testing.T) {
 	assertContents(t, dst, "# Main\n")
 }
 
+func TestGCSGetter_GetGenerationFile(t *testing.T) {
```
sylviamoss, line 92: Nice!

sylviamoss (approved):
LGTM 👍🏼

azr (approved):
LGTM too !

schmichael (approved):
LGTM :tada:

ScottSuarez:
> LGTM too !

Hi @azr I was wondering if you could help me to migrate the test data to the hashicorp controlled bucket? I think it might be best to maintain the data there in the long run? What do you think?

azr:
Hey @ScottSuarez, of course ! Yeah I created it there to be quick. Dm me ?

PR #312, opened 2021-03-16
TestGCSGetter_GetGenerationFile: use our test bucket

azr:
After #310, @ScottSuarez proposed we put the GCP generational content in our bucket so that test things are in one place. 
This PR uses our bucket.

### DIFF
--- a/get_gcs_test.go
+++ b/get_gcs_test.go
@@ -98,7 +98,7 @@ func TestGCSGetter_GetGenerationFile(t *testing.T) {
 
 	// Download
 	err := g.GetFile(
-		dst, testURL("https://www.googleapis.com/storage/v1/go-getter-testcase-data/DO_NOT_DELETE/generation_test.txt#1614317688843055"))
+		dst, testURL("https://www.googleapis.com/storage/v1/go-getter-test/go-getter/versioned.txt#1615905097179533"))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


ScottSuarez (approved):
lgtm :)

get_gcs_test.go
```diff
@@ -108,13 +108,13 @@ func TestGCSGetter_GetGenerationFile(t *testing.T) {
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
-	if string(content) != "a" {
+	if string(content) != "a\n" {
```
radeksimko, line 111: Is there a reason we're adding the newlines to the test data?

radeksimko (approved):
Just one remaining question - Is the access to the bucket documented somewhere internally? No need to discuss where or how (here on GitHub), just checking to make sure it is.

get_gcs_test.go
azr: Nope, I created the file using `echo a > file.txt` and it added a `\n`, and I just felt like this was an okay change.

radeksimko: That's fair. Trailing newline seems like a good practice - I just didn't know if there's anything more to it 😄

azr: Thanks for reviewing !

PR #315, opened 2021-03-31
HttpGetter: Avoid reusing the same http.Request

radeksimko:
This fixes a race condition as demonstrated by the attached test.

`go-cleanhttp` is bumped to `0.5.2` as that brings this patch https://github.com/hashicorp/go-cleanhttp/commit/6d9e2ac5d828e5f8594b97f88c4bde14a67bb6d2 which in combination with the implementation introduces the race condition.

See also https://github.com/golang/go/issues/45311

This will effectively unblock https://github.com/hashicorp/terraform-ls/pull/448 once `terraform-exec` upgrades `go-getter` to a version with this patch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,9 @@ commands:
         type: string
       platform:
         type: string
+      govet:
+        type: string
+        default: ""
     steps:
       - run:
           name: "Run go tests"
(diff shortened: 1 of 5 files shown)


azr (approved):
Nice one, LGTM

PR #300, opened 2021-01-06
Fixes #299

nl-brett-stime:
Does not override GIT_SSH_COMMAND when not needed. See #299 .

PR #316, opened 2021-03-31
Allows alternate profile for S3 refs.

nl-brett-stime:
Allows e.g., Terraform modules to be pulled from a different account than the one where the S3 backend is being initialized.

Some organizations use multiple accounts for security isolation between e.g., dev, test and production environments. Custom modules may be defined in another, distinct account for build / CICD / artifact repository purposes.

By default, the AWS SDK will look for credentials in a section of `~/.aws/credentials` when the `AWS_PROFILE` environment is set with the name of the section. In order to pull modules from one account and apply them another, it would be useful to specify different different profiles for the source and target accounts. This patch allows setting a different environment key (per module reference) to indicate which profile to use for the source account (the application-target account can keep using the default key of `AWS_PROFILE`).

### DIFF
--- a/get_s3.go
+++ b/get_s3.go
@@ -273,15 +273,41 @@ func (g *S3Getter) parseUrl(u *url.URL) (region, bucket, path, version string, c
 		}
 	}
 
+	_, hasProfileEnvKey := u.Query()["aws_profile_env_key"]
 	_, hasAwsId := u.Query()["aws_access_key_id"]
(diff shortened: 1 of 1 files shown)


nl-brett-stime:
FWIW, I think this addresses issues similar to those addressed by #275

nl-brett-stime:
I've tried to use the form to sign the CLA but it says "You have signed the CLA for multiple repositories or organizations"--probably from when I submitted #300 . When I try to load it from this PR, the form is grayed out.

PR #319, opened 2021-04-28
Add insecure flag for http getter.

chrisgilmerproj:
I have a need to use `go-getter` without checking the certificate chain on the http server. To accomplish this I've added a flag named `-insecure` that will allow me to do this. I've tried to add tons of warnings in the code and in the tool to make users aware what a bad idea it is to use this flag and that they are taking risks.

Prior to using this flag I would see:

```text
2021/04/28 14:11:06 Error downloading: Get "https://username:***@hostname": x509: cannot validate certificate for hostname because it doesn't contain any IP SANs
```

After adding the flag I get this message:

```text
2021/04/28 14:37:17 WARNING: Using Insecure TLS transport!
2021/04/28 14:37:17 success!
```

I am unsure how to add a test that shows this without also bootstrapping a server with bad TLS in the tests, but if that's needed I can do it. I've also tried to follow the paradigms in the code but if I've missed something I can fix it.

Finally, I noticed that one of the error messages kept spitting out an unredacted password in a URI. I added that as a security fix here but I'm happy to remove it and submit another PR.

### DIFF
--- a/client_option_insecure.go
+++ b/client_option_insecure.go
@@ -5,7 +5,7 @@ package getter
 // For example, when connecting on HTTPS where an
 // invalid certificate is presented.
 // User assumes all risk.
-// Not all getters have progress support yet.
+// Not all getters have support for insecure mode yet.
 func WithInsecure() func(*Client) error {
 	return func(c *Client) error {
 		c.Insecure = true
--- a/client_option_insecure.go
+++ b/client_option_insecure.go
@@ -1,0 +1,14 @@
+package getter
+
+// WithInsecure allows for a user to avoid
+// checking certificates (not recommended).
+// For example, when connecting on HTTPS where an
+// invalid certificate is presented.
+// User assumes all risk.
+// Not all getters have progress support yet.
+func WithInsecure() func(*Client) error {
+	return func(c *Client) error {
+		c.Insecure = true
+		return nil
+	}
+}
--- a/client.go
+++ b/client.go
@@ -67,6 +67,15 @@ type Client struct {
 	// By default a no op progress listener is used.
 	ProgressListener ProgressTracker
 
+	// Insecure controls whether a client verifies the server's
+	// certificate chain and host name. If Insecure is true, crypto/tls
+	// accepts any certificate presented by the server and any host name in that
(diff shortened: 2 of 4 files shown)


SwampDragons:
@tgross you offered to take a look at this one.

client_option_insecure.go
```diff
@@ -0,0 +1,14 @@
+package getter
+
+// WithInsecure allows for a user to avoid
+// checking certificates (not recommended).
+// For example, when connecting on HTTPS where an
+// invalid certificate is presented.
+// User assumes all risk.
+// Not all getters have progress support yet.
```
tgross, line 8:
```suggestion
// Not all getters have support for insecure mode yet.
```

Looks like this was a copy-paste error from `client_option_progress.go`. But this also points out to me that there's some precedent for having options that don't support all getters, which was one of my initial concerns here. :+1:

tgross (changes requested):
LGTM once that fix in `client_option_insecure.go` is made.

There's a legitimate concern here about whether we really want to support an insecure option, but as long as it's entirely opt-in we're not at risk that our known consumers like TF/Packer/Nomad are going to accidentally misconfigure it. So I'm :+1:

client_option_insecure.go
chrisgilmerproj: I pushed a fix for the above, nice catch. I am not a huge fan of having an option that doesn't work everywhere but once I saw the progress bar I thought it would be ok. Also, there aren't good parallels for `insecure` for some of the other getters and I wasn't sure how to handle that.

chrisgilmerproj:
> There's a legitimate concern here about whether we really want to support an insecure option, but as long as it's entirely opt-in we're not at risk that our known consumers like TF/Packer/Nomad are going to accidentally misconfigure it. So I'm 👍

I stand with you on the concern about making this available. There are definitely modern use cases for servers which are improperly set up however making this available means folks can abuse poorly configured systems without addressing security concerns. Thankfully its opt-in and even if it doesn't get merged folks can still pull the code from this PR if they really need it. 

Thanks for the review, I really appreciate you giving it some attention.

tgross (approved):

PR #321, opened 2021-05-07
http: Initialize http.Request with ctx so cancelation interrupts the request

### LINKED ISSUE #320: context is not respected for HTTP requests
It seems the passed `Ctx` to `getter.Client` is not passed to the `http.Request`:

https://github.com/hashicorp/go-getter/blob/main/get_http.go#L85-L94

I wrote a test for this below. I receive the context.Canceled error, but it still waits for both requests (HEAD + GET) to finish, thus this test takes 6 seconds instead of <1sec.

```go
package cache

import (
	"context"
	"errors"
	"net/http"
	"net/http/httptest"
	"testing"
	"time"

	"github.com/hashicorp/go-getter"
)

func TestHashicorpGoGetter__RespectsContextCanceled(t *testing.T) {
	ctx, cancel := context.WithCancel(context.Background())
	cancel() // cancel immediately

	srv := GivenHTTPServer(t, http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
		t.Logf("Received %s %s", req.Method, req.URL)
		time.Sleep(3 * time.Second)
	}))

	tmpdir := t.TempDir()
	g := getter.Client{
		Ctx:  ctx,
		Src:  srv.URL,
		Pwd:  tmpdir,
		Dst:  "testfile",
		Mode: getter.ClientModeFile,
	}

	t1 := time.Now()
	err := g.Get()
	dur := time.Since(t1)

	if !errors.Is(err, context.Canceled) {
		t.Log("error is not context.Canceled")
		t.Fail()
	}
	if dur > 1*time.Second {
		t.Log("Get() took too long")
		t.Fail()
	}
}

func GivenHTTPServer(t testing.TB, h http.Handler) *httptest.Server {
	srv := httptest.NewServer(h)
	t.Cleanup(func() {
		srv.CloseClientConnections()
		srv.Close()
	})
	return srv
}
```

Obviously it is more realistic to cancel the context during the request. This could be achieved by moving the `cancel()` call into the http handler.

zeisss:
Fixes #320

I tried following the structure in the existing tests. I will create a separate PR for the `v2` branch.

### DIFF
--- a/get_http_test.go
+++ b/get_http_test.go
@@ -1,6 +1,7 @@
 package getter
 
 import (
+	"context"
 	"crypto/sha256"
 	"encoding/hex"
 	"errors"
@@ -404,6 +405,42 @@ func TestHttpGetter_cleanhttp(t *testing.T) {
 	}
 }
 
+func TestHttpGetter__RespectsContextCanceled(t *testing.T) {
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // cancel immediately
+
+	ln := testHttpServer(t)
+
+	var u url.URL
+	u.Scheme = "http"
+	u.Host = ln.Addr().String()
+	u.Path = "/file"
+	dst := tempDir(t)
+
+	rt := hookableHTTPRoundTripper{
+		before: func(req *http.Request) {
+			err := req.Context().Err()
+			if !errors.Is(err, context.Canceled) {
+				t.Fatalf("Expected http.Request with canceled.Context, got: %v", err)
+			}
+		},
+		RoundTripper: http.DefaultTransport,
+	}
+
+	g := new(HttpGetter)
+	g.client = &Client{
+		Ctx: ctx,
+	}
+	g.Client = &http.Client{
+		Transport: &rt,
+	}
+
+	err := g.Get(dst, &u)
+	if !errors.Is(err, context.Canceled) {
+		t.Fatalf("expected context.Canceled, got: %v", err)
+	}
+}
+
 func testHttpServer(t *testing.T) net.Listener {
 	ln, err := net.Listen("tcp", "127.0.0.1:0")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


get_http_test.go
```diff
@@ -404,6 +406,35 @@ func TestHttpGetter_cleanhttp(t *testing.T) {
...
+
+	start := time.Now()
+	err := g.Get(dst, &u)
+	dur := time.Since(start)
+
+	if !errors.Is(err, context.Canceled) {
+		t.Fatalf("expected context.Canceled, got: %v", err)
+	}
+	if dur > 1*time.Second {
+		t.Fatal("Get() took too long")
```
azr, line 434: This test might get flaky with CI tools and I think that the http client will return before actually sending the request, so the err check should be enough IMO.
If you agree then let's remove it ?

zeisss: At least in my production code I got the context.Canceled due to the copy operations later aborting and returning the context.Canceled. 
The important part here is that we use the `ctx` in the `http.Request`. We could write our own mock http.RoundTripper and test the context there, but I am not sure if that isn't too focused.

azr (approved):
Hello there, thanks for opening so quickly, this LGTM, just had a suggestion for tests.

Also, I think you need to sign the cla before we can merge this one 🙂

get_http_test.go
azr: I was trying to say that time-bound testing is not always a good way to test because often: ci tools are shared instances, and sometimes they can just randomly freeze for a (possibly long) moment.

zeisss: > and sometimes they can just randomly freeze for a (possibly long) moment.

Agreed. I will rework the code a bit to explicitly check in the transport or sthg. Stay tuned 👍

azr: But the above tests line #430 sorts of guarantees that this is valid, because the error has to be a `context.Canceled` one.

azr:
No worries, I see only one notification on GH 🙂, good findings !

zeisss:
@azr I have signed the CLA now.

Regarding the tests: I have added a workaround by introspecting the request passed to the transport. If you don't want that nuance I can drop it. Otherwise this should be ready to go. Please take a look.

azr (approved):
Cool beans ! This looks good to me, thanks.

PR #324, opened 2021-05-31
fixes #320: Pass context to NewRequestWithContext so HTTP requests can…

### LINKED ISSUE #320: context is not respected for HTTP requests

zeisss:
Fix for #320 for the `v2` branch. More or less the same test + fix.

### DIFF
--- a/get_http.go
+++ b/get_http.go
@@ -80,7 +80,7 @@ func (g *HttpGetter) Get(ctx context.Context, req *Request) error {
 	req.u.RawQuery = q.Encode()
 
 	// Get the URL
-	httpReq, err := http.NewRequest("GET", req.u.String(), nil)
+	httpReq, err := http.NewRequestWithContext(ctx, "GET", req.u.String(), nil)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


sylviamoss (approved):
LGTM 👍🏼

azr (approved):
Dope, thanks !

PR #325, opened 2021-06-17
CI: bump MSV of go to 1.15.13

tgross:
9e42df5 introduced an `Insecure` flag for the http getter and in its error
reporting relied on the `url.Redacted()` method introduced in golang 1.15
(See https://golang.org/doc/go1.15). This breaks CI, but golang 1.14 is out of
support so our minimum supported version of go should now be on the 1.15.x
series.


notnoop (approved):
LGTM. Good catch.

PR #327, opened 2021-06-24
Honor copy=true in directory mode for local file copies

ebudan:
This looks like an oversight: when getting local files, directory copies always result in a symlink. This may not be desirable in all use cases.

### DIFF
--- a/get_file_unix.go
+++ b/get_file_unix.go
@@ -45,7 +45,11 @@ func (g *FileGetter) Get(dst string, u *url.URL) error {
 		return err
 	}
 
-	return os.Symlink(path, dst)
+	if !g.Copy {
+		return os.Symlink(path, dst)
+	}
+
+	return copyDir(g.Context(), dst, path, false, g.client.umask())
 }
 
 func (g *FileGetter) GetFile(dst string, u *url.URL) error {


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/hashicorp/go-getter
+module github.com/ebudan/go-getter
```
sylviamoss, line 1:
```suggestion
module github.com/hashicorp/go-getter
```

get_file_unix.go
```diff
@@ -41,11 +41,15 @@ func (g *FileGetter) Get(dst string, u *url.URL) error {
 	}
 
 	// Create all the parent directories
-	if err := os.MkdirAll(filepath.Dir(dst), g.client.mode(0755)); err != nil {
+	if err := os.MkdirAll(dst, g.client.mode(0755)); err != nil {
```
sylviamoss, line 44: Why removing the `filepath.Dir`?

sylviamoss (commented):
This is looking good 👍🏼 
I miss a test though. A test similar to:  https://github.com/hashicorp/go-getter/blob/main/get_file_test.go#L129

ebudan:
Added test as requested, and reverted my dir creation change - the reason for that was obscure and not useful here. Module name was changed as I had to use my fork - reverted, will need an additional fork unless this request works for you.

Two realizations during this commit: this leaves the Windows version uncovered (no opportunity here), and there is no support in the CLI tool for hard copy. What would be preferrable for the latter: a command line option (that only applies to copy... ouch), or a url parameter (go-getter file:///tmp/foo?copy=true /tmp/dst ...ouch)?

go.mod
ebudan: See comments below

get_file_unix.go
ebudan: See comments below

Smithx10:
Merge this?

PR #328, opened 2021-06-29
export tar decompressor

schmichael:
Asked for in #19, added in #138, and now exported to match others


dadgar (approved):

ketzacoatl:
Yay! Thanks @schmichael!

PR #331, opened 2021-07-27
Pass signore client ID and secret to releaser

mdeggies:
The release job is failing because the signore client ID and secret have not been set in the release job: https://github.com/hashicorp/go-getter/runs/3167243861. These are passed in to the [setup-signore job](https://github.com/hashicorp/go-getter/commit/d4be3502b44c5acc628bdffdf230189cf5e0136b), but it seems they also need to be made available explicitly in the goreleaser context. This PR is meant to address that.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,8 +67,8 @@ jobs:
           ARTIFACTORY_USER: ${{ secrets.ARTIFACTORY_USER }}
           CIRCLE_TOKEN: ${{ secrets.CIRCLE_TOKEN }}
           SIGNORE_SIGNER: ${{secrets.SIGNORE_SIGNER}}
-          SIGNORE_CLIENT_ID: $${{secrets.SIGNORE_CLIENT_ID}
-          SIGNORE_CLIENT_SECRET: $${{secrets.SIGNORE_CLIENT_SECRET}
+          SIGNORE_CLIENT_ID: ${{secrets.SIGNORE_CLIENT_ID}}
+          SIGNORE_CLIENT_SECRET: ${{secrets.SIGNORE_CLIENT_SECRET}}
       - name: Run clamAV antivirus scanner
         run: sudo clamscan /home/runner/work/$REPO/$REPO/dist/
         env:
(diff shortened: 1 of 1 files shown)


bflad:
Setting the client id and secret in setup-signore should be creating a configuration file which can be used by later signore invocations, however that step is silently failing with:

```
(node:1577) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/home/runner/.signore/config.yaml'
```

We should update the action to fail loudly for those errors (and of course resolve the issue itself 😄 ). A temporary workflow workaround may be to manually `mkdir $HOME/.signore` before setup-signore.

mdeggies:
Ah, thanks Brian! The secrets can also be set as env vars, right? (Though perhaps we'd need to set more than just the client ID + secret?) :-D

milescrabill:
Yes, the secrets can be set as env vars - `SIGNORE_CLIENT_ID`, `SIGNORE_CLIENT_SECRET`, `SIGNORE_SIGNER`, and more: https://github.com/hashicorp/signore/blob/main/docs/CONFIGURATION.md#client

.github/workflows/release.yml
```diff
@@ -67,6 +67,8 @@ jobs:
           ARTIFACTORY_USER: ${{ secrets.ARTIFACTORY_USER }}
           CIRCLE_TOKEN: ${{ secrets.CIRCLE_TOKEN }}
           SIGNORE_SIGNER: ${{secrets.SIGNORE_SIGNER}}
+          SIGNORE_CLIENT_ID: $${{secrets.SIGNORE_CLIENT_ID}
+          SIGNORE_CLIENT_SECRET: $${{secrets.SIGNORE_CLIENT_SECRET}
```
bflad, line 71: Extra `$` and missing `}` 😎 

```suggestion
          SIGNORE_CLIENT_ID: ${{secrets.SIGNORE_CLIENT_ID}}
          SIGNORE_CLIENT_SECRET: ${{secrets.SIGNORE_CLIENT_SECRET}}
```

bflad (approved):
Otherwise this should be good to unblock the release process for now. 👍

.github/workflows/release.yml
mdeggies: Oh geez, thanks for catching this Brian. :D

azr (approved):

mdeggies:
Ok! Merging & running attempt #2. Thanks y'all

PR #333, opened 2021-08-12
s3: Force setting region always in AWS session

mattrobenolt:
I describe the problem in comment in code, but tl;dr is when using `?aws_profile=xxx` method, we never set `region` on the AWS session unless it's defined in the profile itself. This is incorrect behavior since buckets exist in different regions than your profile and are explicitly defined anyways.

Without this, in my case, I was getting a `MissingRegion` error when calling `client.GetObject()`.

### DIFF
--- a/get_s3.go
+++ b/get_s3.go
@@ -176,7 +176,7 @@ func (g *S3Getter) getObject(ctx context.Context, client *s3.S3, dst, bucket, ke
 	return copyReader(dst, resp.Body, 0666, g.client.umask())
 }
 
-func (g *S3Getter) getAWSConfig(region string, url *url.URL, creds *credentials.Credentials) *aws.Config {
+func (g *S3Getter) getAWSConfig(url *url.URL, creds *credentials.Credentials) *aws.Config {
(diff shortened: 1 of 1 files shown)


get_s3.go
```diff
@@ -302,9 +299,18 @@ func (g *S3Getter) newS3Client(
...
+		config := g.getAWSConfig(url, creds)
 		sess = session.New(config)
 	}
 
+	// When using an AWS_PROFILE, the region is coming from your ~/.aws/config
+	// This may not exist in your profile, and it also may differ from the bucket
+	// location, so we need to override and explicitly bind what region we're
+	// intending based on the bucket location otherwise it's possible to get
+	// a MissingRegion error.
+	if region != "" {
```
mattrobenolt, line 311: I didn't want to make this change since I didn't look _that_ deep, but I think we can avoid this check entirely since afaict, this value will always be something. Any detection of region that gets passed in here comes through at least with a fallback of `us-east-1` when there's no region defined in the url itself. But I kept it since it existed in `getAWSConfig` before.

PR #334, opened 2021-08-12
s3: Add progress tracker support to s3 downloads

mattrobenolt:
This was oddly missing entirely and I was real sad. It seems to only be
supported on HTTP right now.

### DIFF
--- a/get_s3.go
+++ b/get_s3.go
@@ -173,7 +173,15 @@ func (g *S3Getter) getObject(ctx context.Context, client *s3.S3, dst, bucket, ke
 		return err
 	}
 
-	return copyReader(dst, resp.Body, 0666, g.client.umask())
+	body := resp.Body
+
+	if g.client != nil && g.client.ProgressListener != nil {
+		fn := filepath.Base(key)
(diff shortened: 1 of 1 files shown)


get_s3.go
```diff
@@ -173,7 +173,15 @@ func (g *S3Getter) getObject(ctx context.Context, client *s3.S3, dst, bucket, ke
 		return err
 	}
 
-	return copyReader(dst, resp.Body, 0666, g.client.umask())
+	body := resp.Body
+
+	if g.client.ProgressListener != nil {
```
azr, line 178: Watch out, the tests paniced here for me:
```suggestion
	if g.client != nil && g.client.ProgressListener != nil {
```

azr (approved):
LGTM after applying my suggestion, thanks

mattrobenolt:
Interesting, I wonder if that case _only_ applies during test suite, since I couldn't find any path where `client` could be `nil` in normal use. Lemme add that check tho.

get_s3.go
mattrobenolt: Added.

azr:
Thanks ! Yes, the tests test things in isolation 🙂 .

PR #339, opened 2021-09-20
add installation procedure by asdf-vm

ryodocx:
Hi,
I created plugin for asdf-vm.
If you like it, add installation procedure by asdf-vm to readme.

FYR:
* asdf-vm: https://asdf-vm.com (not mine)
* plugin: https://github.com/ryodocx/asdf-go-getter
* PR at asdf-vm: https://github.com/asdf-vm/asdf-plugins/pull/481

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,15 +35,24 @@ $ go get github.com/hashicorp/go-getter
 
 go-getter also has a command you can use to test URL strings:
 
+install via `go install`:
 ```
 $ go install github.com/hashicorp/go-getter/cmd/go-getter
 ...
+```
 
-$ go-getter github.com/foo/bar ./foo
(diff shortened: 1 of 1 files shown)


ryodocx:
@nywilken 
Could you help this PR?

CreatorHead (changes requested):
Thanks! This is a small docs-only change and looks like a good candidate to merge.
One request before we merge: since the asdf plugin is community-maintained (not part of this repo), can we tweak the wording to make that explicit and include the plugin URL in the install command?

Something like: “Install via the community-maintained asdf plugin:” and use `asdf plugin add go-getter https://github.com/ryodocx/asdf-go-getter.git` so the instructions are reliable even if the plugin isn’t available via the default asdf plugin index.

Issue #343, opened 2021-12-20
bug(decompress_zip): "runtime error: invalid memory address or nil pointer dereference"


joaoReigota1:
When running decompress on a protected zip such as [protected-AES-256.zip](https://github.com/hashicorp/go-getter/files/7744725/protected-AES-256.zip) which has a password: `123`,  go-getter returns a panic in decompress_zip.go:77

![image](https://user-images.githubusercontent.com/74597872/146756327-1fd3615e-e267-4f94-8a3d-b7557abff848.png)

Maybe to fix this panic:
```
srcF, err := f.Open()
if err != nil {
  if srcF != nil {
    srcF.Close()
  }
  return err
}
```

PR #344, opened 2021-12-20
fix(decompresser_zip): fix panic when decompressing protected zip file (#343)

joaoReigota1:
Signed-off-by: João Reigota <joao.reigota@checkmarx.com>

Added check if file is nil before trying to close preventing unwanted panic


azr (approved):
Good catch !

Issue #217, opened 2019-10-13
Support files signed and encrypted with PGP


kshpytsya:
It would convenient to support decryption and signature verification of files signed with PGP. There appears to exist a golang [library](https://godoc.org/golang.org/x/crypto/openpgp) for that.
My specific use case involves Nomad job [artifacts](https://www.nomadproject.io/docs/job-specification/artifact.html). Artifact checksum is unknown at the time of job creation as artifacts can change without Nomad job modification. And, of course, it would be useful to have encryption as well. I do realize that is is possible to have `go-getter` just download the encrypted file, and do decryption/verification manually, but that just doesn't scale well to multiple jobs using different Docker images.

PR #350, opened 2022-01-11
Ignore comments in checksum files

azr:
This is a tiny simple step towards #217

### DIFF
--- a/checksum.go
+++ b/checksum.go
@@ -307,6 +307,11 @@ func (c *Client) checksumFromFile(ctx context.Context, checksumURL string, check
 // for BSD type sums parseChecksumLine guesses the hashing algorithm
 // by checking the length of the checksum.
 func parseChecksumLine(line string) (*FileChecksum, error) {
+	switch line[0] {
+	case '#', '/', '-':
+		return nil, nil // skip
+	}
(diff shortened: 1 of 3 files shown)


nywilken (approved):
This looks good to me. I was curious on what would happen if the test file looked like the example in the packer issue. More specifically containing lines that are not checksums or comments but it looks like `parseChecksumLine` would return an error and the caller would continue. So this looks good. 

```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256
```

checksum.go
```diff
@@ -307,6 +307,11 @@ func (c *Client) checksumFromFile(ctx context.Context, checksumURL string, check
 // for BSD type sums parseChecksumLine guesses the hashing algorithm
 // by checking the length of the checksum.
 func parseChecksumLine(line string) (*FileChecksum, error) {
+	switch line[0] {
+	case '#', '/', '-':
+		return nil, nil // skip
+	}
+	//TODO: this function will fail if we pass a checksum for a path with spaces
```
sylviamoss, line 314: I'm not sure if I understood this one 👀

sylviamoss (approved):
👍🏼

azr:
Thx !

PR #351, opened 2022-01-30
Append tar and ova decompressor in v2

UnConTesteur:
Hello,

I try to implement this issue https://github.com/hashicorp/packer-plugin-qemu/issues/7.
The plugin use `go-getter` v2 for decompress archive, however `tar` or `ova` are unsupported in this version.

So I append `tar` support. I cherry-pick the commit 6a2a0516f4917e30a3e3e00593081a6b1273a8ab and append two new entries in the decompressor map.

### DIFF
--- a/decompress.go
+++ b/decompress.go
@@ -23,6 +23,7 @@ type Decompressor interface {
 var Decompressors map[string]Decompressor
 
 func init() {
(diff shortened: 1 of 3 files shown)


azr (approved):
LGTM, good catch, merci ! 🙂

PR #352, opened 2022-02-10
[v2] update s3 gcs and cmd to use most recent v2 module

azr:


azr:
I would be nice to not have a circular dependency here, will try to think of something later on. But it will be a breaking change because of the `Request` and `Mode` types

PR #363, opened 2022-05-20
Update default Getters with known defaults

nywilken:
* By default XTerrformGet header support should be disable; clients
  needing supporting this header should overwrite the default getters
  with a custom HttpGetter

* Add default timeouts for HTTP Read Requests; including HEAD requests

### DIFF
--- a/get.go
+++ b/get.go
@@ -19,6 +19,7 @@ import (
 	"os/exec"
 	"regexp"
 	"syscall"
+	"time"
 
 	cleanhttp "github.com/hashicorp/go-cleanhttp"
 )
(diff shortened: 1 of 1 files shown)


get.go
```diff
@@ -70,7 +71,10 @@ var DefaultClient = &Client{
 
 func init() {
 	httpGetter := &HttpGetter{
-		Netrc: true,
+		Netrc:                 true,
+		XTerraformGetDisabled: true,
+		HeadFirstTimeout:      10 * time.Second,
```
picatz, line 76: 🤔 Maybe we should go with `DoNotCheckHeadFirst`, to avoid the initial `HEAD` request altogether.

```suggestion
		DoNotCheckHeadFirst:   true,
```

What do you think?

sylviamoss (approved):
👍🏼  LGTM!

get.go
nywilken: Packer uses the default Getter settings with a few additional Getters. I updated them to match what we plan to use in Packer.  That said, I think Packer could take advantage of the Range header. So I opted for the HeadFirstTimeout over the disabling it entirely. Does that sound reasonable to you?

nywilken: Given the need of the changes. I went ahead and merged as is. If after working through Packer further we find we can disable entirely. We can revisit. But sticking with this default for now.

PR #364, opened 2022-05-20
Update X-Terraform-Get error messaging to match v1

nywilken:
All in the title.

### DIFF
--- a/get_http_test.go
+++ b/get_http_test.go
@@ -48,7 +48,7 @@ func TestHttpGetter_header(t *testing.T) {
 
 	// Get it, which should error because it uses the file protocol.
 	err := g.Get(ctx, req)
-	if !strings.Contains(err.Error(), "") {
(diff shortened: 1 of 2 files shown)


get_http_test.go
```diff
@@ -48,7 +48,7 @@ func TestHttpGetter_header(t *testing.T) {
 
 	// Get it, which should error because it uses the file protocol.
 	err := g.Get(ctx, req)
-	if !strings.Contains(err.Error(), "download not supported for scheme") {
+	if !strings.Contains(err.Error(), "") {
```
picatz, line 51: 🤔 I _think_ we're missing the argument to `strings.Contains` here?

```suggestion
	if !strings.Contains(err.Error(), "no getter available for X-Terraform-Get source protocol") {
```

sylviamoss (approved):
👍🏼  Nice!

PR #370, opened 2022-06-07
Check detectors and relative paths before proto from X-Terraform-Get

### LINKED ISSUE #368: no getter available for X-Terraform-Get source protocol
### Terraform Version

```
Terraform v1.2.2
```

### Terraform Configuration Files
```terraform
module "cloudfront" {
  source = "https://api.example.com/module"
  [...]
}
```

### Expected Behavior
```terraform
Downloading https://api.example.com/module cloudfront...
- cloudfront in .terraform/modules/cloudfront
```

### Actual Behavior
```terraform
Could not download module "cloudfront" (cloudfront.tf:1) source code from
│ "https://api.example.com/module": error
│ downloading 'https://api.example.com/module':
│ no getter available for X-Terraform-Get source protocol:
│ "github.com/terraform-aws-modules/terraform-aws-cloudfront?ref=v2.9.3"
```

### Steps to Reproduce
```terraform
terraform init
```

### Additional Context
We are using custom API to return modules locations and versions, it is working fine with terraform 1.2.1 but it is not working anymore with terraform 1.2.2

jbardin:
When checking for an unexpected protocol switch via X-Terraform-Get, we
first must run the configured detectors and check for a relative request
path, since X-Terraform-Get is not required to return a valid URL.

Fixes #368

### DIFF
--- a/get_http.go
+++ b/get_http.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


get_http.go
```diff
@@ -293,50 +294,66 @@ func (g *HttpGetter) Get(dst string, u *url.URL) error {
 	// If there is a subdir component, then we download the root separately
 	// into a temporary directory, then copy over the proper subdir.
 	source, subDir := SourceDirSubdir(source)
-	if subDir == "" {
-		var opts []ClientOption
+	if subDir != "" {
+		// We have a subdir, time to jump some hoops
+		return g.getSubdir(ctx, dst, source, subDir)
```
nywilken, line 299: I was a little confused about the call to getSubdir the first time I read this code. Is the source being used here the value returned from the X-Terraform-Get header? 

If so, would we not want to run the protocol switch check before returning?  Or will this be handled in the final call to Get that is made in the getSubdir function?

get_http.go
```diff
@@ -293,50 +294,66 @@ func (g *HttpGetter) Get(dst string, u *url.URL) error {
...
+		// next Get anyway.
+		source, err = Detect(source, g.client.Pwd, g.client.Detectors)
+		if err != nil {
+			return err
+		}
+
+		protocol := ""
+		// X-Terraform-Get allows paths relative to the previous request too,
+		// which won't have a protocol.
+		if !relativeGet(source) {
```
nywilken, line 318: Nicely done!

get_http.go
```diff
@@ -293,50 +294,66 @@ func (g *HttpGetter) Get(dst string, u *url.URL) error {
...
+		// not necessarily return a valid URL. We can replace the source string
+		// here, since the detectors would have been called immediately during the
+		// next Get anyway.
+		source, err = Detect(source, g.client.Pwd, g.client.Detectors)
+		if err != nil {
+			return err
+		}
+
+		protocol := ""
+		// X-Terraform-Get allows paths relative to the previous request too,
```
nywilken, line 316: For my knowledge in the case of a relative path request how does go-getter know the url used in the previous request? I don't see it being passed into the next Get request.

get_http_test.go
```diff
@@ -584,6 +584,47 @@ func TestHttpGetter__XTerraformGetDisabled(t *testing.T) {
...
+
+func (testCustomDetector) Detect(src, _ string) (string, bool, error) {
+	if strings.HasPrefix(src, "custom|") {
+		return "http://" + src[7:], true, nil
+	}
+	return "", false, nil
+}
+
+// test a source url with no protocol
+func TestHttpGetter__XTerraformGetDetected(t *testing.T) {
```
nywilken, line 597: Should we add a test for relative paths?

nywilken (approved):
The changes look good to me. I have a couple of questions on the handling of sub directory and relative path requests. I'm approving in advance in case my questions are just due to my gaps of knowledge around go-getter.

 But feel free to ping me again for another review if the questions result in a change.

get_http.go
jbardin: That's a good question. While the PR looks big because I dropped a level of indentation by returning early, I left the overall logic the same. I think you may be right that a subdir would bypass the the protocol checks since it goes on to call `Get` just like below -- I'll follow up with a refactor of `getSubdir` so we can pass in the same options

get_http.go
jbardin: I don't think relative paths actually work in this mode, but I added the check for completeness, and so that we can return any intended error later, rather than a more confusing `no getter available` message.

get_http.go
jbardin: The new proto checks were bypassed by the addition of a subdirectory component to the url. I refactored this a bit to move the `getSubdir` call after the checks, and pass in the same options we would use for `Get`.

get_http_test.go
nywilken: Got it!

nywilken (approved):
Thanks for answering my questions. The updated check for subDir looks good to me. I pinged @picatz just for visibility as it changes how the getter logic was understood at first.

get_http.go
```diff
@@ -293,50 +294,79 @@ func (g *HttpGetter) Get(dst string, u *url.URL) error {
...
+	// This prevents all default getters from being allowed when only using the
+	// HttpGetter directly. To enable protocol switching, a client "wrapper" must
+	// be used.
+	if g.client == nil {
+		switch {
+		case subDir != "":
+			// If there's a subdirectory, we will also need a file getter to
+			// unpack it.
+			opts = append(opts, WithGetters(map[string]Getter{
+				"file":  new(FileGetter),
```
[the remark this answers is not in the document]
picatz: Thank you for the ping!

The intent for this original check was to totally prevent the `HttpGetter` from allowing cross-protocol switching unless the getter was already supported by a `Client`. I understand the reasoning for the `FileGetter` with a subdirectory component, but I _hate_ this is allowed at all. It has the potential to be dangerous, and generally not desirable from a security standpoint.

Thankfully, I don't think it would cause problems for Nomad, who explicitly disable the `FileGetter` because it's dangerous in their context. They use a `Client` directly. But I could see this causing problems in other contexts. Likely outside of HashiCorp. 

I would much rather this be a new option on the `HttpGetter`, rather than be implicitly allowed by default. Maybe something like:

```go
httpGetter := &HttpGetter{
  AllowFileGetterForSubDirectoryComponent: true,
}
```

What do you think? Would that be an acceptable solution?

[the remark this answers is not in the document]
jbardin: To be clear, this is more limiting than the current release, since `getSubdir` falls back to all the default getters. We can't really change this in version 1, since it's a breaking change. While `X-Terraform-Get` probably shouldn't have been so liberal in spec, the protocol does define any valid go-getter source string is a valid value, including subdirectories. I think the only valid way to avoid this is to disable `X-Terraform-Get`, which we have an option for already.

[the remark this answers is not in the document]
picatz: To be clear `getSubdir` was falling back to all default getters when using the `HttpGetter` directly?

Aside from that clarification: I think I agreed with you, that disabling `X-Terraform-Get` would generally be the "right" solution, as opposed to an additional option.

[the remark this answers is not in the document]
jbardin: Right now these checks are only run if `subDir == ""`, otherwise we just call [`getSubDir`](https://github.com/hashicorp/go-getter/blob/main/get_http.go#L339), which runs [`Get`](https://github.com/hashicorp/go-getter/blob/c5e9d081082f01577c64ae9af69f06e20a1e3bd0/get_http.go#L495) with the default options, bypassing the forced `http`+`https` getters entirely. This change now at least will force the `http`+`https` getters with subdirectories too.

picatz (approved):

eastebry (approved):

Issue #374, opened 2022-07-13
Feature: Handle GCS Service Account Key Directly through Environment Variable


GoodmanBen:
This issue originated while trying to access modules stored in GCS from within Terraform Cloud.

### Current state:
The [recommended ways](https://www.terraform.io/language/modules/sources#gcs-bucket) of authenticating with GCS to use as a module source are not possible when running Terraform through Terraform Cloud.

#### Set the GOOGLE_OAUTH_ACCESS_TOKEN environment variable to a raw Google Cloud Platform OAuth access token.
- Not possible since the token expires after a set amount of time, constantly generating and placing a new token value into Terraform Cloud is not a clean option to pursue.

#### Enter the path of your service account key file in the GOOGLE_APPLICATION_CREDENTIALS environment variable.
- Not possible to store the service account key as a file within Terraform Cloud without checking the file into the Terraform configuration/source code as plain text. This is not a viable option security-wise.

#### If you're running Terraform from a GCE instance, default credentials are automatically available. See [Creating and Enabling Service Accounts](https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances) for Instances for more details.
- Would like to run Terraform on Terraform Cloud, not GCE, so this is not really an option.

#### On your computer, you can make your Google identity available by running `gcloud auth application-default login`.
- Running this command is not possible within a SaaS Terraform Cloud workflow.

### Desired state:
User can specify an environment variable `GOOGLE_CREDENTIALS`, which contains the contents of the service account key, as a means of authentication.

This would match the [GOOGLE_CREDENTIALS](https://support.hashicorp.com/hc/en-us/articles/4406586874387-How-to-set-up-Google-Cloud-GCP-credentials-in-Terraform-Cloud) environment variable recommended by HashiCorp for letting Terraform Cloud control GCP resources.

PR #375, opened 2022-07-13
Feature: Get from GCS, Service Account Key Directly through ENV Variable

GoodmanBen:
Resolves https://github.com/hashicorp/go-getter/issues/374.

If packaged into a new release and then incorporated into Terraform, this would allow a single [environment variable](https://support.hashicorp.com/hc/en-us/articles/4406586874387-How-to-set-up-Google-Cloud-GCP-credentials-in-Terraform-Cloud) within Terraform Cloud to grant access to GCP resources as well as allow module downloads from GCS storage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,7 +364,19 @@ Some examples for these addressing schemes:
 
 #### GCS Authentication
 
-In order to access to GCS, authentication credentials should be provided. More information can be found [here](https://cloud.google.com/docs/authentication/getting-started)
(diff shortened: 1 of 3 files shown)


GoodmanBen:
Hi @nywilken, would you be the right HashiCorp team member to ask to review this PR? Looks like you had been the most consistently active member in this repo recently but I'm not sure if someone else would be better to ask.

This new functionality will un-block our team's full transition to Terraform Cloud as our IaC management platform.

PR #376, opened 2022-07-19
parse checksum even when no EOL is found (#285)

nickryand:
I am pulling over a fix from the main branch to v2. I have run into an issue using packer trying to build a NixOS qemu image. The published NixOS sha files don't end in a newline, they end in an EOF.


nickryand:
@nywilken @claire-labry think this PR can be merged in? The fix already exists in V1.

PR #378, opened 2022-08-30
Update S3 URL in README

twelvelabs:
The bucket access info located in the previous URL has been moved to a separate page.


kmoe (approved):
Thanks

PR #384, opened 2022-10-10
Update aws sdk version

### LINKED ISSUE #383: S3 module downloads doesn't work with AWS IRSA
[IAM roles for service accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) is a feature of EKS to provide IAM role creds to kubernetes pods. The [aws-sdk version (v1.15.78) used in the current version of go-getter](https://github.com/hashicorp/go-getter/blob/293f9e2372e5cd5ff8cdf7b6977b861122837292/go.mod) doesn't support IRSA in its default credential chain. Versions after v1.23.13 are supported ([source](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-minimum-sdk.html)). To overcome this problem, a recent version of aws-sdk should be used.

The use case: We bumped into this problem while using terraform (via atlantis) in a kubernetes pod.

Jukie:
Closes #383


kmoe (approved):
Thanks

Jukie:
@kmoe could we have a new v1 release cut so we can bring these changes into upstream terraform?

PR #379, opened 2022-09-20
Migrate to GHA

claire-labry:
This PR officially migrates `go-getter` to GHA from CircleCI. It's intention is to work the same as the CircleCI tests but in GHA format. 

Note: After working on updating the AWS/GCP tests, it was concluded that the older go versions (1.14.15 and 1.15.3) were blocking the GCP tests from passing. The issue was that go1.17 is required for the dependent GCS libraries and we concluded that the removal of the older go versions was the logical path to move forward.

PR #388, opened 2022-10-20
Migrate go-getter v2 to GHA

claire-labry:
This PR migrates `v2` of `go-getter` to GHA from CircleCI.


get_hg_test.go
```diff
@@ -126,6 +127,11 @@ func TestHgGetter_HgArgumentsNotAllowed(t *testing.T) {
 		t.Log("hg not found, skipping")
 		t.Skip()
 	}
+
+	if runtime.GOOS == "windows" {
+		t.Log("skipping for Windows OS for now")
```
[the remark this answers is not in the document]
claire-labry: 👍🏻

get_hg_test.go
```diff
@@ -126,6 +127,11 @@ func TestHgGetter_HgArgumentsNotAllowed(t *testing.T) {
 		t.Log("hg not found, skipping")
 		t.Skip()
 	}
+
+	if runtime.GOOS == "windows" {
+		t.Log("skipping on Windows OS for now")
+		t.Skip()
+	}
```
picatz, line 134: Just to clarify, why are we skipping this test for Windows? 🤔

claire-labry: great question (I could've definitely added some sort of comment to clarify this bit) When @nywilken and I were pairing, we kept running issues with getting the windows test to pass with the error below for `hg` -- I'm not entirely sure I understand why, but Wilken may be able to provide more context than I can.

https://github.com/hashicorp/go-getter/actions/runs/3299997899/jobs/5444000536
Screen Shot 2022-10-26 at 11 07 00 AM

claire-labry:
@picatz would you be able to re-review this PR as I've added several things since your last review: [this commit here](https://github.com/hashicorp/go-getter/pull/379/commits/bad826d9c545eb209425c991226c487b0093641d) -- we need to allow the file to be transported globally due to the CVE linked in the comments. I also added [comments](https://github.com/hashicorp/go-getter/pull/379/commits/47ecbb0e0aae24d0304b71bea95a0f89153e0d2f) to the GCS and S3 tests for external contributors. Once I have your final review and approval, I will merge this in.

get_hg_test.go
nywilken: 👋 great question. In pairing with @claire-labry I came to believe that these tests were not being previously tested on Windows via CircleCI. So I suggested skipping for now to get the GHA work in place. 

Looking at the v1 code and this code further it appears that only one of the test cases is failing specifically the `"arguments allowed in destination"` test cases, which is not a test case within the v1 code base. 

The test case has a touch command, which I don't think is available on Windows. Maybe that's the issue. An alternative to skipping this whole test on Windows is breaking out the failing case and just skipping that one on Windows for now. I don't have the cycles to dive into it further at this time. So I'm proposing the Windows skip.

picatz: Totally fair, thank you both for the additional context. It's not preferred, but totally reasonable if you don't have additional cycles. My only worry is that I could imagine this being fairly confusing in a few months time, or for an external contributor.

Perhaps we should capture this in an issue or add a comment to explain why it's only "for now".

picatz:
```suggestion
	if runtime.GOOS == "windows" {
	        // Please refer to https://github.com/hashicorp/go-getter/pull/388/files#r1005819432
	        // for more context why we are temporarily skipping Windows OS.
		t.Log("skipping on Windows OS for now")
		t.Skip()
	}
```

claire-labry: Great suggestion-- will add this.

PR #400, opened 2022-11-30
[v2] update s3, gcs and cmd to use most recent v2 module version

nywilken:


nywilken:
I opened a PR to have the acceptance test run. Since it is all green I will merge and tag each of the submodules for release.

PR #408, opened 2023-01-24
remove codesign entirely from go-getter

claire-labry:
Removing codesigning from this project as users only and primarily use the library instead of the go binaries.


mdeggies (approved):
Looks good! Thanks Claire

PR #414, opened 2023-02-09
Port Decompression bomb security changes from v1

### LINKED ISSUE #407: Vulnerable to Zip bombs.
Zip bombs or compression bombs are trick archives that are custom structured to generate Terabytes of garbage (or zeros) from a small archive. It turns out go-getter's decompression options and detectors are susceptible to this. This is behaviour by design but it might be a good idea if go-getter verified an archive's size will fit in storage before decompression.  This doesn't just apply to zip files but also xz, zstd, etc. Downloading a zip bomb today is actually detected and block by Chrome and some browsers to prevent you from accidentally filling your partition.

A crafty Nomad job of type "system" could instantly bring down an entire Nomad DC if the artifact is downloaded to all nodes and expanded to fill all storage.  This pseudojob is missing the `exec` driver.
```
job "bomb" {
  type = "system"
  datacenters = ["dc1"]
  group "artifacts" {
    task "server" {
      artifact {
        source      = "https://www.bamsoftware.com/hacks/zipbomb/zblg.zip"
        destination = "/tmp/some-directory"
        options {
          checksum = "md5:207b597f03033b2e0644bbbc29f04053"
        }
      }
    }
  }
}
```
Beware zip bombs and please use caution if trying to reproduce this:
https://www.bamsoftware.com/hacks/zipbomb/

```
$ ~/go/bin/go-getter https://www.bamsoftware.com/hacks/zipbomb/zblg.zip /tmp/
2023/01/19 13:11:00 Error downloading: write /tmp/1: no space left on device
```

nywilken:
This PR aims to fix #407 (reserved as `CVE-2023-0475`), by introducing "(de)compression" bomb mitigation options to the various [decompressors](https://github.com/hashicorp/go-getter/blob/d229395f5a7d9f36340f313c7130ae3852cedc1a/decompress.go#L14) provided by this package. Specifically, `FileSizeLimit` and `FilesLimit`.

* `FileSizeLimit` limits the size of a decompressed file, or limits the total size of all decompressed files if dealing with an archive.
* `FilesLimit` limits the number of files that are allowed to be decompressed from an archive.

These changes were ported from v1. 

In addition to the security changes support for ZTSD as a supported decompressor was also ported.


gcs/get_gcs.go
```diff
@@ -21,6 +21,12 @@ type Getter struct {
 	// Timeout sets a deadline which all GCS operations should
 	// complete within. Zero value means no timeout.
 	Timeout time.Duration
+
+	// FileSizeLimit limits the size of an single
+	// decompressed file.
+	//
+	// The zero value means no limit.
+	FileSizeLimit int64
```
picatz, line 29: This was probably left-over from a previous patch I shared. This shouldn't be required.
```suggestion
```

picatz (changes requested):

gcs/get_gcs.go
nywilken: Thanks for calling that out. I saw it was not being used so was hoping the review would shed some light on it. I'll remove.

picatz (approved):
LGTM 👍

PR #413, opened 2023-02-09
v1: decompressors: add LimitedDecompressors helper

shoenig:
This PR adds helper function LimitedDecompressors(filesLimit, fileSizeLimit)
for creating the same suite of decompressors as Decompressors, but all configured
with the given filesLimit and fileSizeLimit (where applicable).

PR #415, opened 2023-02-10
v2: decompressors: add LimitedDecompressors helper

nywilken:
This change ports the helper function LimitedDecompressors(filesLimit, fileSizeLimit)
for creating the same suite of decompressors as Decompressors, but all configured
with the given filesLimit and fileSizeLimit that was introduced in
go-getter v1.

Changes ported from #413


picatz (approved):
LGTM 👍

PR #417, opened 2023-02-13
Update Go-Getter test workflow to run on pull requests

nywilken:
Currently test for v2 run for authors who have push access to the go-getter repo. This change updates the test action to run for all open pull-request to the v2 branch, as well as on pushes to the v2 branch.

### DIFF
--- a/.github/workflows/go-getter.yml
+++ b/.github/workflows/go-getter.yml
@@ -1,6 +1,13 @@
 name: go-getter
 
-on: [push]
+on: 
+  push:
+    branches:
+      - v2
+  pull_request:
+    branches:
+      - v2
+
 
 env:
   TEST_RESULTS_PATH: /tmp/test-results


.github/workflows/go-getter.yml
```diff
@@ -1,6 +1,11 @@
 name: go-getter
 
-on: [push]
+on: 
+  push:
+    branches:
+      - v2
+  pull_request:
```
lbajolet-hashicorp, line 7: Out of curiosity what does this empty section imply?

lbajolet-hashicorp: I presume it means that it runs on all PRs pushing to v2?

lbajolet-hashicorp (approved):
LGTM

.github/workflows/go-getter.yml
nywilken: Ah good call out. This will actually run for all pull requests. I can update for any pull_requests against v2

nywilken: Updated

PR #424, opened 2023-03-10
Remove `bomb.zip` test file to stop anti-virus noise

### LINKED ISSUE #419: Anti-viruses report the bomb.zip test file as malicious
Hello,
This is for sure related to https://github.com/hashicorp/go-getter/issues/407 but the introduction of a zip bomb file into the repository generates alerts in EDR/Antivirus: https://www.virustotal.com/gui/file/f1dc920869794df3e258f42f9b99157104cd3f8c14394c1b9d043d6fcda14c0a

This will either create troubles for repository users, or may even prevent the go-getter project from working properly (either because the AV will delete the test file and so, tests will fail or because the AV will block the access to the file when running the tests).

May you remove this file from the repository, or "encrypt" it so AV won't complain? I think even a simple XOR with a hard coded "password" inside the test would be enough for the AVs to not complain about the bomb.zip file (you may upload it to VT and check if any AV still consider the zip as malicious)

Thanks

picatz:
This PR aims to fix #419. The `bomb.zip` file was useful to verify the zip bomb protections worked properly, but is not required to verify out mitigations work. Including a real zip bomb lead to confusion and reports of anti-virus causing issues for users. Instead, a zip file is created inline with the test before verifying the decompression limit works.

### DIFF
--- a/decompress_zip_test.go
+++ b/decompress_zip_test.go
@@ -1,6 +1,10 @@
 package getter
 
 import (
+	"archive/zip"
+	"bytes"
+	"io/ioutil"
+	"log"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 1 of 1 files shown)


picatz:
cc @nywilken we probably need to get this same change landed in v2 sometime in the near future.

nywilken:
> cc @nywilken we probably need to get this same change landed in v2 sometime in the near future.

Sounds good thanks for the detailed description. I like the approach. This change can be easily cherry-picked into v2 once merged.

nywilken (approved):
LGTM

nywilken:
> cc @nywilken we probably need to get this same change landed in v2 sometime in the near future.

Sounds good thanks for the detailed description. I like the approach. This change can be easily cherry-picked into v2 once merged.

PR #426, opened 2023-03-16
[v2] update s3, gcs and cmd to use most recent v2 module version

nywilken:
- [v2] update s3, gcs and cmd to use most recent v2 module version


lbajolet-hashicorp (approved):
LGTM!

Issue #422, opened 2023-03-02
GitHub https downloads break when there is a / in the branch name.


lazzurs:
The issue appears to be with the logic here on splitting the URL based on / here

https://github.com/hashicorp/go-getter/blob/0edab85348271c843782993345b07b1ac98912e6/detect_github.go#L25

When using this in Terragrunt the following works

    source = "github.com/xxx/infrastructure-modules//dev-dns-zone?ref=v0.0.1"

But this does not

    source = "github.com/xxx/infrastructure-modules//dev-dns-zone?ref=feat/PTP-1841"

The error in Terragrunt clearly shows the branch name is having the part after / chopped off.

```
ERRO[0001] 1 error occurred:
* error downloading 'https://github.com/xxx/infrastructure-modules.git?ref=feat': /opt/homebrew/bin/git exited with 1: error: pathspec 'feat' did not match any file(s) known to git

ERRO[0001] Unable to determine underlying exit code, so Terragrunt will exit with error code 1
```

Issue raised in Terragrunt for this is https://github.com/gruntwork-io/terragrunt/issues/2473

cwyl02:
Seeing this issue in our code base as well

PR #430, opened 2023-04-11
fix: downloading github url with / in branch name

cwyl02:
This fixes https://github.com/hashicorp/go-getter/issues/422

### DIFF
--- a/detect_github.go
+++ b/detect_github.go
@@ -23,13 +23,17 @@ func (d *GitHubDetector) Detect(src, _ string) (string, bool, error) {
 }
 
 func (d *GitHubDetector) detectHTTP(src string) (string, bool, error) {
-	parts := strings.Split(src, "/")
-	if len(parts) < 3 {
+	parts := strings.Split(src, "?")
+	if len(parts) > 2 {
+		return "", false, fmt.Errorf("there is more than 1 '?' in the URL")
+	}
+	hostAndPath := parts[0]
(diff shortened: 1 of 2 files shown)


cwyl02:
Hi @picatz @claire-labry , please take a look. We use go-getter internally and merging this will fix https://github.com/hashicorp/go-getter/issues/422 thus help getting our internal CI passing.

cwyl02:
any updates on this PR?

CreatorHead (changes requested):
Thanks for working on this, the overall approach makes sense.
Splitting the source on `?` before path parsing correctly fixes the issue where `/ ` inside query parameters (e.g. branch names like `feature/foo`) were being misinterpreted as part of the repo path.

I have one small concern before merging:

**Handling multiple `?` characters**
Right now the code errors if `strings.Split(src, "?")` results in more than two parts. In practice, only the first `?` is a delimiter in a URL; additional `?` characters can legally appear inside the query string as data. Previously, these cases would likely have worked, so this introduces a stricter behavior than before.

Would it make sense to switch to `strings.SplitN(src, "?", 2)` and drop the `len(parts) > 2` check? That would preserve the fix while avoiding a potential regression.

**Tests**
The added test is helpful. It would be great to also include a case that mirrors the original failure more directly, e.g. a URL with `ref=feature/foo` (optionally with a subdirectory as well), to clearly lock in the intended behavior.

With those tweaks, I think this change would be in good shape to merge.

PR #433, opened 2023-04-22
Make addAuthFromNetrc ignore ENOTDIR errors

adrian-bl:
The function already returns early if the specified 'netrc' configuration points to a non existing file, but currently returns an error if the OS reports ENOTDIR:

This will happen if the $HOME directory of the current user points to a file instead a directory - something eg. 'void linux' does for user 'nobody':

```
$ grep nobody /etc/passwd
nobody:x:99:99:Unprivileged User:/dev/null:/bin/false
```

go-getter then attempts to open `/dev/null/.netrc` which fails with ENOTDIR - something that should just be treated the same way as a non existing file (in this case)

### DIFF
--- a/netrc.go
+++ b/netrc.go
@@ -5,6 +5,7 @@ import (
 	"net/url"
 	"os"
 	"runtime"
+	"syscall"
 
 	"github.com/bgentry/go-netrc/netrc"
 	"github.com/mitchellh/go-homedir"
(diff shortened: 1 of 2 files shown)


adrian-bl:
Friendly ping: would be nice to see this fixed upstream

crw:
Hi @adrian-bl, I ran this past the Terraform, Nomad, and Product Security teams, and got sign-off to merge. Thanks for the submission!

PR #435, opened 2023-05-10
Add support for NTLM into HTTP getter

0x4c6565:
Adds support for NTLM authentication into HTTP getter


crw:
Thanks for this submission! Is there a bug where the use case for this change is discussed? Is it targeting a particular product (e.g. Terraform, Nomad, etc)?

PR #459, opened 2023-10-09
don't change GIT_SSH_COMMAND when there is no sshKeyFile

jbardin:
The reason for `setupGitEnv` is to add a `-i keyfile` argument to `GIT_SSH_COMMAND`. 
If there is no key file, we do not need to modify the environment at all.

The problem with #300 was that it always reset `GIT_SSH_COMMAND` even when there was no `keyfile` to add, and lost the original command string.


apparentlymart (approved):
This looks like a plausible fix to me!

PR #465, opened 2023-10-21
Fix typo in README.md

tired-engineer:


crw:
Thank you for this submission. We generally do not accept typo-style fixes to files to reduce merge conflicts and `git blame` confusion, but the next time someone is editing this file, we will fix this typo. Thanks again for this submission!

PR #483, opened 2024-04-12
Escape User Provided Strings to Git

mcollao-hc:
Escapes user provided strings to Git by not allowing it to be used an a Git switch/option.


tgross (approved):
LGTM! ~(For V1, this will need a V2 PR as well.)~ _edit: nope, v1 only_

PR #488, opened 2024-04-18
gcs/v2: bump cloud.google.com/go/storage to 1.34.0

lbajolet-hashicorp:
The storage library that gcs depends on transitively imports grpc in version 1.50.0, which is vulnerable to GHSA-m425-mq94-257g.

While this is a server-side vulnerability, therefore the package is not directly vulnerable (nor its clients), this dependency still causes advisories to be produced against this package, so we update those now.


.github/workflows/go-getter.yml
```diff
@@ -19,9 +19,9 @@ jobs:
     strategy:
       matrix:
         go-version:
-          - 1.17
-          - 1.18
           - 1.19
+          - "1.20"
```
JenGoldstrich, line 23: Should this be a string?
```suggestion
          - 1.20
```

lbajolet-hashicorp: Yes as otherwise YAML "simplifies" it as `1.2` (float automatic casting), so this one has to be quoted.
The others don't get converted so they are fine without quoting.

JenGoldstrich: We talked about this in slack but just for Git context I think you should add a comment here explaining that, gotta love YAML though 🤣

JenGoldstrich (approved):
LGTM, we should document the removal of go 1.17 and 1.18 support in the release notes but it makes sense that we need to remove those as the tests no longer work on newer dependency versions.

nywilken (approved):
LGTM

PR #489, opened 2024-05-20
[v2] update s3, and gcs to use most recent v2 module version

nywilken:


JenGoldstrich (approved):
LGTM, I think you need to run go mod tidy though based on the CI failures

Issue #479, opened 2024-04-01
Unable to download group gitlab repository files 


denis256:
Noticed that `go-getter` fail to download files from gitlab repositories which are located in subgroups:

```
(go-getter_1.7.3_linux_amd64)

$ go-getter "github.com/denis256/s3.git//module2" /tmp/qwe
2024/04/01 18:01:56 success!

$ go-getter "gitlab.com/denis256/s3.git//module2" /tmp/qwe
2024/04/01 18:10:40 success!

$ go-getter "gitlab.com/denis25-tests/dc/platform/terraform-modules/s3.git//module2" /tmp/qwe
2024/04/01 18:02:19 Error downloading: error downloading 'https://gitlab.com/denis25-tests/dc.git': /usr/bin/git exited with 128: Cloning into '/tmp/getter72444115/temp'...
remote: The project you were looking for could not be found or you don't have permission to view it.
fatal: repository 'https://gitlab.com/denis25-tests/dc.git/' not found

```

References:
https://github.com/denis256/s3
https://gitlab.com/denis256/s3
https://gitlab.com/denis25-tests/dc/platform/terraform-modules/s3

gmaghera:
Here is the related documentation https://docs.gitlab.com/ee/user/group/subgroups/

Subgroups can be "Be nested up to 20 levels."

gmaghera:
Could you associate PR https://github.com/hashicorp/go-getter/pull/494 with this issue?

gmaghera:
bump

Issue #470, opened 2024-02-15
support URL to local files related to git top-level 


ChaosEternal:
when refering to a local file, it will be easier to use a syntax like "//path/to/file" than use "../../../path/to/file", here "//" can resolved as `git rev-parse --show-toplevel`.

PR #494, opened 2024-06-03
fix: allow nested repos for gitlab.com

tvizor:
Address #479, #470 (for gitlab.com)

As per GitLab documentation, repositories can be nested under projects and subgroups: https://docs.gitlab.com/ee/user/group/subgroups/

MR is to remove the limit of 3 components of the path (gitlab.com/username/repo) and allow referencing paths to directories inside the repo.

Parse must be performed on the whole URL before evaluating and adding the .git suffix to the repo name to keep the ref query in its original format (?ref=master vs %3Fref=master if parsed as a part of the path).


crw:
I am not sure why I am only now seeing this notification in GitHub, but I will raise this in triage. I find this comment interesting, https://github.com/gruntwork-io/terragrunt/issues/3031#issuecomment-2214576840, will see what is underneath that difference. To set expectations, there is a strong headwind to getting PRs reviewed and merged into this repo due to multiple products using it and an unwillingness to inadvertently add new bugs. In any case, thanks for the submission!

mazzy89:
I'm not using Terraform. Instead, I'm utilizing go-getter in Helmfile, which does not fetch repositories in GitLab sub-groups due to this issue. If I understand correctly, go-getter v2 addresses this issue, right?

PR #498, opened 2024-06-28
Backport fix from https://github.com/hashicorp/go-getter/pull/497 to v2

nywilken:
Recreate git config during update to prevent git config alteration

Related to: #497

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Dockerfile to create a go-getter container with smbclient dependency that is used by the get_smb.go tests
-FROM golang:1.15
+FROM golang:1.19.13
 
 COPY . /go-getter
 WORKDIR /go-getter
--- a/get_git.go
+++ b/get_git.go
@@ -122,7 +122,7 @@ func (g *GitGetter) Get(ctx context.Context, req *Request) error {
 		return err
 	}
 	if err == nil {
-		err = g.update(ctx, req.Dst, sshKeyFile, ref, depth)
+		err = g.update(ctx, req.Dst, sshKeyFile, ref, req.URL(), depth)
 	} else {
 		err = g.clone(ctx, sshKeyFile, depth, req)
 	}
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Dockerfile to create a go-getter container with smbclient dependency that is used by the get_smb.go tests
-FROM golang:1.15
+FROM golang:1.19.13
```
dduzgun-security, line 2: Would using a newer version of go based of an alpine one (eg: `1.22.4-alpine3.20`) work here?

```suggestion
FROM golang:1.22.4-alpine3.20
```

dduzgun-security (commented):
Thanks a lot for the backport 🙌 

Should we also backport the fix we did [here](https://github.com/hashicorp/go-getter/commit/268c11cae8cf0d9374783e06572679796abe9ce9) too (maybe in another PR)

Dockerfile
nywilken: Not sure we can try. I opted for 1.19 because our tests run with 1.19 and 1.20. If it is not relevant to the fix. I would rather open a separate PR updating versions of Go used on CI.

nywilken:
> Thanks a lot for the backport 🙌
> 
> Should we also backport the fix we did [here](https://github.com/hashicorp/go-getter/commit/268c11cae8cf0d9374783e06572679796abe9ce9) too (maybe in another PR)

When presented with this [PR](https://github.com/hashicorp/go-getter/commit/268c11cae8cf0d9374783e06572679796abe9ce9) in Slack @sylviamoss and @mcollao-hc validated that v2 is not susceptible to the vulnerability because v2 does not have a function called `findRemoteDefaultBranch` nor does it execute a command similar to `exec.CommandContext(ctx, "git", "ls-remote", "--symref", "--", u.String(), "HEAD")`

Please advise if your testing is showing different results.

eastebry (approved):

dduzgun-security (approved):

Dockerfile
dduzgun-security: That sounds good 👍  we can tackle it in a separate PR

dduzgun-security:
👋 @nywilken, this looks good thanks a lot for taking care of the backport

nywilken:
> 👋 @nywilken, this looks good thanks a lot for taking care of the backport

I work on getting this released by EoD tomorrow the latest. Thanks for review and gentle nudge.

PR #503, opened 2024-07-24
[v2] Bumps go-getter/v2/cmd/go-getter to use go-getter/gcs/v2 and go-getter/s3/v2 @v2.2.3

nywilken:


lbajolet-hashicorp (approved):
LGTM!

PR #504, opened 2024-10-09
Clean up git repo on disk when the ref checkout fails

james-warren0:
If the `ref` checkout during a call to `clone()` fails, this ensures the cloned repo on disk will get cleaned up.


NodyHub (approved):
I checked the other calls of [`checkout`](https://github.com/hashicorp/go-getter/blob/768e85f2e16bfc1082f3d16bbda885c1821311c8/get_git.go#L176) and looks like this is the only location where we need to handle this failure.

LGTM!

PR #511, opened 2025-01-07
Update Checkout Actions

ryuichi1208:
ref: https://github.com/actions/checkout/commit/11bd71901bbe5b1630ceea73d27597364c9af683


crw:
Hi @ryuichi1208, apologies, but we do not accept changes to the build process from outside contributors. I appreciate the contribution and hopefully this dependency will be updated with the next go-getter release.

PR #513, opened 2025-01-09
IND-1804 Bump up dependencies to remediate vulnerabiities

mohanmanikanta2299:
Bumping up the dependency versions to remediate the vulnerabilities in go-getter repo for v1. 
Also removed go version 1.17 from workflow file as it doesn't work with updated dependencies.


mukeshjc (approved):
LGTM

PR #528, opened 2025-04-28
fix: url redact of multiple sshkey

dduzgun-security:
Ref: https://hashicorp.atlassian.net/browse/SECVULN-14559

### DIFF
--- a/url.go
+++ b/url.go
@@ -9,7 +9,8 @@ import "net/url"
 // which is like url.String but replaces any password with "redacted".
 // Only the password in u.URL is redacted. This allows the library
 // to maintain compatibility with go1.14.
-// This port was also extended to redact SSH key from URL query parameter.
+// This port was also extended to redact all "sshkey" from URL query parameter
+// and replace them with "redacted".
 func RedactURL(u *url.URL) string {
 	if u == nil {
 		return ""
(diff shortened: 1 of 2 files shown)


url.go
```diff
@@ -20,8 +21,12 @@ func RedactURL(u *url.URL) string {
...
 	}
 	q := ru.Query()
-	if q.Get("sshkey") != "" {
-		q.Set("sshkey", "redacted")
+	if q.Has("sshkey") {
+		values := q["sshkey"]
+		for i := range values {
+			values[i] = "redacted"
+		}
+		q["sshkey"] = values
```
abhijeetviswa, line 29: Question: Is this reassignment required?

url_test.go
```diff
@@ -83,7 +83,51 @@ func TestRedactURL(t *testing.T) {
...
+			want: "ssh://git@github.com/hashicorp/go-getter-test-private.git?sshkey=redacted",
+		},
+		{
+			name: "multiple SSH keys in URL query parameter",
+			url: &url.URL{
+				Scheme:   "ssh",
+				User:     url.User("git"),
+				Host:     "github.com",
+				Path:     "repo.git",
+				RawQuery: "sshkey&sshkey=secretkey",
```
abhijeetviswa, line 95: Suggestion: Can this be `sshkey=secretkey1&sshkey=secretkey2`? Right now, it's the same as the test `multiple SSH keys with empty and non-empty values`.

abhijeetviswa (changes requested):

url.go
dduzgun-security: Ah good catch, it's indeed not needed. Removed in [80a526c](https://github.com/hashicorp/go-getter/pull/528/commits/80a526cd1b22f0b770dfd158557341f890f3b4a7)

url_test.go
dduzgun-security: Adjusted the tests in [80a526c](https://github.com/hashicorp/go-getter/pull/528/commits/80a526cd1b22f0b770dfd158557341f890f3b4a7) also added one with multiple ssh key values which was missing

abhijeetviswa (approved):
LGTM!

PR #529, opened 2025-05-28
IND-3728 enabling dependabot

KaushikiAnand:
As per the memo: https://go.hashi.co/memo/eng-004,
Enabling the dependabot for identifying and updating the security and version updates in the repositories rather handling it manually.

### DIFF
--- a/.github/dependabot.yaml
+++ b/.github/dependabot.yaml
@@ -8,12 +8,29 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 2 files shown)


.github/dependabot.yaml
```diff
@@ -0,0 +1,16 @@
+version: 2
```
abhijeetviswa, line 4: Please add copyright headers.

mohanmanikanta2299 (approved):

Issue #538, opened 2025-08-07
File download is skipped when headResp.ContentLength is 0


martijnvdp:
HTTP File download is skipped when headResp.ContentLength returns 0

when we access our private terraform module registry hosted on a gitlab deployment behind an entra application proxy the header response returns a contentlength of 0 and than we get an empty archive file (tgz) with terraform that is using go-getter to download the terraform modules.
and get the error from terraform/go-getter:  
`Error opening a gzip reader for C:\...\..\AppData\Local\Temp\getter1099997155\archive: EOF`
 like mentioned here https://github.com/hashicorp/go-getter/issues/537

After investigation we see that the entra application proxy returns a content length of 0 and when doing it from our on prem environment without the proxy in between we do get a content length greater than zero and the file got correctly downloaded
looks like the azure application proxy is setting the content length to 0

its fixed for us when changing the contentlenght condition to > 0
https://github.com/hashicorp/go-getter/blob/37130302313c9294df898ac96e2565a65369ec68/get_http.go#L446

like this

```go 
if headResp.Header.Get("Accept-Ranges") == "bytes" && headResp.ContentLength > 0 {
```

PR #539, opened 2025-08-07
fix: #538 http file download skipped if headResp.ContentLength is 0

martijnvdp:
We have an issue with terraform not downloading our private modules from gitlab when we do it from a location that uses the azure entra application proxy in between.

After investigation it seems that the response.contentlength is 0 in the header and the download gets skipped
it presumes the file already exists and continues to try to extract the empty archive

https://github.com/hashicorp/go-getter/blob/e70221100018573cdc74411c95c19b2a372f6728/get_http.go#L446

changing the condition from `headResp.ContentLength >= 0` to 
`headResp.ContentLength > 0 ` resolves the issue for us

seems that azure proxy doesnt set the content length correctly

## PCI review checklist

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.

### DIFF
--- a/get_http.go
+++ b/get_http.go
@@ -443,7 +443,7 @@ func (g *HttpGetter) GetFile(dst string, src *url.URL) error {
 			if headResp.StatusCode == 200 {
 				// If the HEAD request succeeded, then attempt to set the range
 				// query if we can.
-				if headResp.Header.Get("Accept-Ranges") == "bytes" && headResp.ContentLength >= 0 {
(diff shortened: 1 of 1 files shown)


CreatorHead:
Some servers/proxies return `Content-Length: 0` for HEAD even when the file isn’t empty, which causes go-getter to incorrectly skip the download.
Treating 0 as “unknown size” avoids incorrect `resume/skip` decisions and ensures the file is actually fetched.

PR #541, opened 2025-08-12
SECVULN-24969 -  resolve improper validation CVE-2025-22868

ritikrajdev:
## Task
- resolve improper validation CVE-2025-22868
- updates version to 1.23 for go as the oauth package CVE has been resolved in version 0.27 which requires this go version.
- Please note the tests were failing because of a few internal changes in go 1.23 for windows regarding symlinks.
- Due to these changes a minor logic change has been introduced 
refer to - https://pkg.go.dev/os@master, https://pkg.go.dev/path/filepath@master

NOTE: Update in go version in ci and go.mod as well.

## PCI review checklist

Only way is to revert the pr. Should be merged by squashing for easier revert.

Can impact windows user a bit.


util_test.go
```diff
@@ -32,7 +31,7 @@ func tempEnv(t *testing.T, k, v string) func() {
 // tempFileContents writes a temporary file and returns the path and a function
 // to clean it up.
 func tempFileContents(t *testing.T, contents string) (string, func()) {
-	tf, err := ioutil.TempFile("", "getter")
+	tf, err := os.CreateTemp("", "getter")
```
CreatorHead, line 34: * Replacing deprecated **ioutil** with **os**.CreateTemp, os.MkdirTemp, os.ReadFile, os.WriteFile. This is fine for Go ≥1.17 and is harmless behavior-wise.

go.mod
```diff
@@ -5,22 +5,41 @@ require (
...
+	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
+	google.golang.org/appengine v1.6.7 // indirect
+	google.golang.org/genproto v0.0.0-20230410155749-daa745c078e1 // indirect
 	google.golang.org/grpc v1.56.3 // indirect
 	google.golang.org/protobuf v1.33.0 // indirect
 	gopkg.in/cheggaaa/pb.v1 v1.0.27 // indirect
 )
 
-go 1.13
+go 1.23.0
```
CreatorHead, line 45: CI/release workflow updates target Go 1.23, which aligns with the oauth2 fix notes (jws fix landed around that ecosystem).

**Dev-announcement:** https://groups.google.com/g/golang-dev/c/9lIKDd3Ff1s?utm_source=chatgpt.com&pli=1
**Official Vul Record:** https://pkg.go.dev/vuln/GO-2025-3488?utm_source=chatgpt.com

copy_dir.go
```diff
@@ -26,7 +27,11 @@ func copyDir(ctx context.Context, dst string, src string, ignoreDot bool, disabl
 	var err error
 	resolved, err := filepath.EvalSymlinks(src)
 	if err != nil {
-		return err
+		if runtime.GOOS == "windows" {
```
CreatorHead: **Why was this change made?**

Right now, the code tries to resolve symlinks when copying directories.
On Linux/macOS, if it can’t resolve them, it fails - which is safe.

On Windows, the new change says: “if resolving symlinks fails, just use the original path anyway.”

It looks to me like this is working as a bypass, but please correct me if I’m wrong. If the safety check fails, we shouldn’t just skip past it. Either the validation should be strict, or the process should fail in a clear way.

.github/workflows/go-getter.yml
```diff
@@ -91,8 +90,7 @@ jobs:
     strategy:
       matrix:
         go-version:
-          - 1.18
-          - 1.19
+          - 1.23
```
CreatorHead, line 93: I see that the workflow now only tests on Go 1.23, and support for Go 1.18/1.19 was removed. This effectively drops CI coverage for older Go versions.

Is this intentional?

get_file_test.go
```diff
@@ -27,7 +28,8 @@ func TestFileGetter(t *testing.T) {
 	if err != nil {
 		t.Fatalf("err: %s", err)
 	}
-	if fi.Mode()&os.ModeSymlink == 0 {
+
+	if runtime.GOOS != "windows" && fi.Mode()&os.ModeSymlink == 0 {
```
CreatorHead: The assertion now skips the symlink check on Windows. Why? 

It looks like the Windows symlink check was just skipped to make the tests pass. That feels risky. Need some explanation here.

util_test.go
ritikrajdev: Yes it is harmless behaviour wise. Ideally this should not have caused any tests to fail. But upgrading to go version 1.23 was the major change over here that led to failing of test cases for windows.

go.mod
ritikrajdev: Yes, it is as expected to upgrade to go version 1.23.0

.github/workflows/go-getter.yml
ritikrajdev: Yes, this change is intentional. The required Go version has been updated to 1.23 in go.mod due to the upgrade of the oauth dependency to version 0.27.0. As a result, it's necessary to align the CI and builds with Go 1.23 or later. We can certainly add Go 1.24 to the matrix for testing as well. However, support for older versions like 1.18 and 1.19 is no longer viable, as they fail linting and other checks.

get_file_test.go
ritikrajdev: Resolving it as found a better solution that will be checking the same for windows but with both the conditions i.e. for a symlink and junction.

copy_dir.go
ritikrajdev: Have added strict validations as mentioned.

.github/workflows/go-getter.yml
CreatorHead: Thanks for clarification

copy_dir.go
```diff
@@ -16,6 +17,26 @@ func mode(mode, umask os.FileMode) os.FileMode {
...
+	}
+
+	// On Windows with Go 1.23+, EvalSymlinks may not resolve junctions
+	fi, statErr := os.Lstat(path)
+	if statErr != nil {
+		return "", err // fallback to original error
+	}
+
+	if fi.Mode()&os.ModeSymlink != 0 || fi.Mode()&os.ModeIrregular != 0 {
+		return path, nil
```
CreatorHead, line 33: Just adding a note here for visibility - maybe someone with more context can clarify, but here are my observations:

* If the file is a symlink --> `return path, nil` (pretending it’s fine).
* If the file is irregular (which on Windows usually means a junction / reparse point) → return path, nil.
* If it’s a normal file/dir → also return path, nil.

In all cases, if **EvalSymlinks** failed, the function just hands back the original path with nil error.

This feels risky to me - it might be safer to fail or add stricter checks instead of treating all these cases as safe

.github/workflows/go-getter.yml
```diff
@@ -12,8 +12,7 @@ jobs:
...
```
pkazmierczak, line 15: could we maybe bump up to 1.25 already since that's the latest version available?

copy_dir.go
```diff
@@ -16,6 +17,26 @@ func mode(mode, umask os.FileMode) os.FileMode {
...
+	resolved, err := filepath.EvalSymlinks(path)
+	if err == nil || runtime.GOOS != "windows" {
+		return resolved, err
+	}
+
+	// On Windows with Go 1.23+, EvalSymlinks may not resolve junctions
+	fi, statErr := os.Lstat(path)
+	if statErr != nil {
+		return "", err // fallback to original error
+	}
```
kkavish, line 30: off the top of my head - should the 2 errors be joined `errors.Join(error...)`? or at least be logged? In the interest of not losing any errors.

copy_dir.go
```diff
@@ -16,6 +17,26 @@ func mode(mode, umask os.FileMode) os.FileMode {
 	return mode & ^umask
 }
 
+func safeEvalSymlinks(path string) (string, error) {
```
tgross, line 20: There's nothing in the PR description that explains why we're doing this, just a vague reference to stdlib libraries. It doesn't appear to be at all related to  CVE-2025-22868  which is an oauth library bug. We shouldn't be pushing in unrelated work into a security PR; if there's another CVE to cover that should be its own PR.

.github/workflows/go-getter.yml
tgross: This is a matrix, so we want the currently supported versions: 1.24 and 1.25.

copy_dir.go
kkavish, line 20: The way I understand this is:
1. Resolve symlinks to the actual path
2. Keep others, including windows junctions as is. i.e. junctions won't be resolved.

If my understanding is right, won't this introduce a new SECVULN?
The whole point of resolving symlinks and junctions is to prevent a lower-privileged user from creating a symlink/junction to a privileged item and then have a field day with it from a security POV. The fact that we use this in Terraform and Nomad makes me more worried.

https://msrc.microsoft.com/blog/2025/06/redirectionguard-mitigating-unsafe-junction-traversal-in-windows/

Adding @james-warren0 as well so that he can share his views on this.

get.go
```diff
@@ -84,12 +84,13 @@ func init() {
 // src is a URL, whereas dst is always just a file path to a folder. This
 // folder doesn't need to exist. It will be created if it doesn't exist.
 func Get(dst, src string, opts ...ClientOption) error {
-	return (&Client{
+	result := (&Client{
```
kkavish, line 87: Is this really required, can we revert back to the good olden code?

kkavish (commented):
@ritikrajdev Appreciate the work here! While this looks good to me I have a little concern around how windows junctions are being handled. I have made a small comment about that and looped in @james-warren0 from Sec Team to have a look as well.

Thanks!

copy_dir.go
tgross: @dduzgun-security has spent a _lot_ of time with Nomad and go-getter tracking down this kind of problem so he might have more thoughts. He'll probably suggest that we look at using `os.Root` here to avoid TOCTOU issues too.

james-warren0: I agree. I think `os.Root` is the correct way to address this issue and not miss edge cases. `os.Root` was introduced in go 1.24, so we would need to bump one more version to utilize it.

dduzgun-security: Yes, we definitely should be cautious here. I would strongly recommend working on the bump (without the code changes on how we deal with symlinks) and having another work item / task to explore migrating to os.Root.

.github/workflows/go-getter.yml
dduzgun-security: Yep, the stable versions of go are `1.24.X` and `1.25.X` https://go.dev/dl/ meaning that `1.23.X` won't get updates. We should fix this.

.github/workflows/go-getter.yml
dduzgun-security: Similar comment as https://github.com/hashicorp/go-getter/pull/541/files#r2294091217, we should rely on the stable versions of go which are 1.24 and 1.25.

.github/workflows/release.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - name: Setup go
         uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
         with:
-          go-version: '^1.15'
+          go-version: '^1.23'
```
dduzgun-security, line 25: Similar comment as https://github.com/hashicorp/go-getter/pull/541/files#r2294091217, we should rely on the stable versions of go which are 1.24 and 1.25.

copy_dir.go
ritikrajdev: Hi Tim, the version upgrade was required as minimum supported version for oauth 0.27.0 (patched version) is go version 1.23

ritikrajdev:
using dependabot pr for the same.

PR #557, opened 2025-09-08
[fix] Windows junction handling

YakDriver:
## PCI review checklist

- [x] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [x] If applicable, I've documented the impact of any changes to security controls.

## Fix Windows Test Coverage and Implement Cross-Platform Symlink Resolution

This PR addresses two key areas to improve the robustness and test coverage of go-getter on Windows:

1. Removes Windows test skips for critical file and HTTP getter functionality
2. Implements comprehensive symlink resolution with Windows junction point support

Previously, several important tests were being skipped on Windows, reducing test coverage and potentially hiding Windows-specific issues.

### Cross-Platform Symlink Resolution

Implemented new symlink resolution infrastructure to handle the complexities of Windows filesystem links:
* resolveSymlinks(src string) (string, error) windows/unix versions

All previously skipped Windows tests now pass


symlinks_windows.go
```diff
@@ -0,0 +1,147 @@
...
+	// Return the original EvalSymlinks error
+	if junctionErr != nil {
+		return "", fmt.Errorf("failed to resolve symlinks (checking whether junction point: %s): %w", junctionErr, err)
+	}
+	return "", fmt.Errorf("failed to resolve symlinks: %w", err)
+}
+
+// isWindowsJunctionPoint uses Windows API to reliably detect junction points
+// by checking the reparse point tag specifically.
+func isWindowsJunctionPoint(path string) (bool, error) {
```
ritikrajdev, line 76: How is this implementation different from checking for `fi.Mode & os.ModeIrregular` ?

symlinks_windows.go
```diff
@@ -0,0 +1,147 @@
...
+
+	// Path exists but EvalSymlinks failed on Windows
+	// Check what type of path this is
+
+	// Case 1: Check if this is a junction point
+	var isJunction bool
+	var junctionErr error
+	if isJunction, junctionErr = isWindowsJunctionPoint(src); junctionErr == nil && isJunction {
+		// Confirmed junction point - return cleaned path
+		return filepath.Clean(src), nil
```
ritikrajdev: it will not give us the actual resolved path over here after following the junction link. it would only be reformatting the path in src instead.

symlinks_windows.go
```diff
@@ -0,0 +1,147 @@
...
+		// Return the cleaned path, plain and true.
+		return filepath.Clean(src), nil
+	}
+
+	// If we get here, it's some other type of special file/link that we don't understand
+	// Return the original EvalSymlinks error
+	if junctionErr != nil {
+		return "", fmt.Errorf("failed to resolve symlinks (checking whether junction point: %s): %w", junctionErr, err)
+	}
+	return "", fmt.Errorf("failed to resolve symlinks: %w", err)
```
ritikrajdev, line 71: Would like to appreciate the logic & Edge cases that has been thought over here.

get_file_test.go
```diff
@@ -29,12 +29,26 @@ func TestFileGetter(t *testing.T) {
...
-		if err != nil {
-			t.Fatalf("err: %s", err)
+	// Verify it's a symlink on Unix or junction point on Windows
+	fi, err := os.Lstat(dst)
+	if err != nil {
+		t.Fatalf("err: %s", err)
+	}
+
+	if runtime.GOOS == "windows" {
+		isJunction, junctionErr := isWindowsJunctionPoint(dst)
```
ritikrajdev, line 39: add in a few unit test cases for isWindowsJunctionPoint.

get_http_test.go
```diff
@@ -142,11 +141,6 @@ func TestHttpGetter_meta(t *testing.T) {
 }
 
 func TestHttpGetter_metaSubdir(t *testing.T) {
-	// Skip this test on Windows due to file:// URL subdirectory resolution issues
-	if runtime.GOOS == "windows" {
-		t.Skip("Skipping meta subdir test on Windows due to file:// URL path resolution issues")
-	}
```
ritikrajdev, line 148: good job in removing these if cases over here and handling it in the implementation, hence making it more abstract.

symlinks_windows.go
ritikrajdev: wanted to clarify, how are we going to handle if it's directly a mount point. Os.ModeIrregular as of now on windows represents a file that has been unknown. Can be a "Mount point" or a "Junction". We should ideally add in a check for mount point over here as well.



& just a suggestion, might check this one out later, if we don't have any public package available for this change happening in go 1.23 related to evalsymlink, should we create a go package and distribute it with others?

YakDriver: ## How is this implementation different from checking for `fi.Mode & os.ModeIrregular`?

The key differences are **precision** and **reliability**:

### `os.ModeIrregular` Limitations:
- **Too broad**: `os.ModeIrregular` is a catch-all flag that indicates "unknown file type" on Windows
- **Ambiguous**: It can represent junction points, mount points, symbolic links, or other reparse points
- **No distinction**: You can't tell what specific type of reparse point you're dealing with

### Our Windows API Implementation:
- **Specific detection**: Directly queries the reparse point tag via `FSCTL_GET_REPARSE_POINT`
- **Precise identification**: Distinguishes between `IO_REPARSE_TAG_MOUNT_POINT` (junctions) and `IO_REPARSE_TAG_SYMLINK` (symlinks)
- **Reliable**: Uses the same mechanism Windows itself uses internally

```go
// This tells us "something unusual" but not what
if fi.Mode() & os.ModeIrregular != 0 {
    // Could be junction, mount point, symlink, or other reparse point
}

// This tells us specifically "this is a junction point"
reparseTag := *(*uint32)(unsafe.Pointer(&buffer[0]))
isJunction := reparseTag == IO_REPARSE_TAG_MOUNT_POINT // 0xA0000003
```

## Handling Mount Points vs Junction Points

You're absolutely right about the distinction! The current implementation focuses on junction points, but we should also handle volume mount points:

### Current Implementation:
- Only checks for `IO_REPARSE_TAG_MOUNT_POINT` (0xA0000003) - which is actually **junction points**
- Despite the confusing name, `IO_REPARSE_TAG_MOUNT_POINT` refers to directory junctions, not volume mounts

### What We Should Add:
```go
const (
    IO_REPARSE_TAG_MOUNT_POINT = 0xA0000003  // Junction points
    IO_REPARSE_TAG_SYMLINK     = 0xA000000C  // Symbolic links  
    // We should also handle:
    // Volume mount points (different from junctions)
    // Other reparse point types as needed
)
```

We could extend the function to return more specific information:
```go
type ReparsePointType int
const (
    NotReparsePoint ReparsePointType = iota
    JunctionPoint
    VolumeMount
    SymbolicLink
    // ... other types
)

func getWindowsReparsePointType(path string) (ReparsePointType, error)
```

## Go 1.23 Package Distribution Suggestion

Great point about creating a standalone package! This would be valuable because:

### Benefits:
- **Broader ecosystem impact**: Other projects face similar Windows symlink/junction issues
- **Focused testing**: Dedicated package allows more comprehensive edge case testing
- **Version independence**: Doesn't require Go 1.23+ to get improved Windows path handling
- **Community contribution**: Could become the de facto solution for Windows path resolution

### Potential Package Structure:
```
github.com/hashicorp/go-winpath
├── junction.go      // Junction point detection
├── symlink.go       // Enhanced symlink resolution  
├── mount.go         // Volume mount point handling
├── reparse.go       // General reparse point utilities
└── resolve.go       // Cross-platform path resolution
```

### API Design:
```go
// Cross-platform enhanced path resolution
func ResolvePath(path string) (string, error)

// Windows-specific utilities
func IsJunctionPoint(path string) (bool, error)
func IsVolumeMount(path string) (bool, error) 
func GetReparsePointInfo(path string) (*ReparseInfo, error)
```

This would allow go-getter to use the package while also benefiting the broader Go community dealing with Windows filesystem complexities.

symlinks_windows.go
YakDriver: Yes, good catch. I'll fix this.

symlinks_windows.go
YakDriver: The logic handles the **"fallthrough" error case** when all other resolution attempts have failed. Here's the detailed breakdown:

### The Logic Flow Context:

```go
// We get here when:
// 1. filepath.EvalSymlinks(src) initially failed (line ~18)
// 2. Path exists (os.Lstat and os.Stat succeeded)  
// 3. It's NOT a junction point (junction check failed/returned false)
// 4. It's NOT a regular directory (regular dir check failed)

// If we get here, it's some other type of special file/link that we don't understand
// Return the original EvalSymlinks error  
if junctionErr != nil {
    return "", fmt.Errorf("failed to resolve symlinks (checking whether junction point: %s): %w", junctionErr, err)
}
return "", fmt.Errorf("failed to resolve symlinks: %w", err)
```

### Edge Cases Being Handled:

#### 1. **Junction Point Detection Failed (`junctionErr != nil`)**
```go
if junctionErr != nil {
    return "", fmt.Errorf("failed to resolve symlinks (checking whether junction point: %s): %w", junctionErr, err)
}
```

**When this happens:**
- Path exists and `EvalSymlinks` failed
- Our junction point detection itself failed (API error, permissions, etc.)
- We can't determine if it's a junction or not

**Why return both errors:**
- `junctionErr`: Shows what went wrong during our junction detection
- `err`: The original `EvalSymlinks` error that started this whole fallback process
- Combined context helps debugging: "EvalSymlinks failed AND we couldn't even check if it's a junction"

#### 2. **Unknown Reparse Point Type (`junctionErr == nil`)**
```go
return "", fmt.Errorf("failed to resolve symlinks: %w", err)
```

**When this happens:**
- Path exists and `EvalSymlinks` failed  
- Junction detection succeeded (no error) but returned `false`
- It's not a regular directory
- **It's some other type of reparse point or special file**

**Examples of "other" types:**
- Volume mount points (`IO_REPARSE_TAG_VOLUME_MOUNT`)
- Symbolic links with weird attributes  
- Deduplication reparse points
- Cloud storage reparse points (OneDrive, etc.)
- Custom reparse point types

### The Defensive Programming Approach:

This logic embodies **conservative error handling**:

1. **Try the standard approach** (`EvalSymlinks`)
2. **Handle known special cases** (junction points, regular dirs)  
3. **Gracefully fail on unknown cases** rather than returning potentially wrong paths

### Why Not Try to Handle More Cases?

The code deliberately **doesn't** try to resolve every possible reparse point type because:

- **Safety**: Wrong resolution could break functionality
- **Complexity**: Each reparse type has different resolution logic
- **Scope**: The primary goal was fixing junction points specifically
- **Fallback**: Better to fail explicitly than return incorrect paths

This is a solid **"fail fast, fail clearly"** approach that provides good error context for debugging while avoiding the complexity of handling every possible Windows filesystem edge case.

get_file_test.go
YakDriver: Added

.gitignore
```diff
@@ -1 +1,75 @@
...
+go.work
+
+# Build artifacts
+/bin/
+/build/
+/dist/
+
+# Project-specific binaries
 cmd/go-getter/go-getter
+go-getter
```
ritikrajdev, line 27: just want to confirm if we're ignoring go-getter binary or a directory named go-getter over here?
any specific reason why we're ignoring this>?

ritikrajdev (approved):

.gitignore
YakDriver: there to keep the compiled go-getter binary out of version control

PR #560, opened 2025-09-22
Ensure temporary directory symlinks do not cause errors

cam-stitt:
There was a [recent change](https://github.com/hashicorp/go-getter/commit/87541b2501c00df5eaedea6acc61a2a4a4efa5b7) that introduced a check to avoid symlinks inside git sub-directories. This is a good change, however there were two problems at play.

First, there was no test for a "successful" get of a sub-directory within the git client. Second, some operating systems, when creating a temporary directory via `os.Mkdirtemp`, return a symlink. This meant that *no* git sub-directory gets would work due to an unintentional symlink.

This change has two commits. The first is a test to showcase the error, the other replaces the superfluous `go-safetemp` by bringing the tiny library internal, and evaluating any symlinks upon temporary directory creation. There is no other HashiCorp tool utilising the `go-safetemp` package, and it simply creates an indirect dependency for minimal utility code.

## PCI review checklist

### DIFF
--- a/client.go
+++ b/client.go
@@ -13,7 +13,6 @@ import (
 	"strings"
 
 	urlhelper "github.com/hashicorp/go-getter/helper/url"
-	safetemp "github.com/hashicorp/go-safetemp"
 )
 
 // ErrSymlinkCopy means that a copy of a symlink was encountered on a request with DisableSymlinks enabled.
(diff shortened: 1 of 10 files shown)


eastebry:
@dduzgun-security and @nodyhub - would you mind taking a look at this?

sonamtenzin2 (approved):

PR #564, opened 2025-10-22
Remove failed decompressed files

allisonlarson:
When untarring a tarball, attempt to create files initially with the
umask file permissions and delete the file if it fails to copy
correctly. This prevents a partially written file from remaining on disk
with permissions that didn't include the umask.

Adds a unit test and a invalid tarball (created by truncating tarball containing a large
file generated from /dev/urandom), which fails to decompress but errors
partially written.

## PCI review checklist

### DIFF
--- a/get_file_copy.go
+++ b/get_file_copy.go
@@ -47,7 +47,8 @@ func copyReader(dst string, src io.Reader, fmode, umask os.FileMode, fileSizeLim
 
 	_, err = io.Copy(dstF, src)
 	if err != nil {
-		// Remove the file in case of partial write
+		// Close & remove the file in case of partial write
+		_ = dstF.Close()
 		_ = os.Remove(dst)
 		return err
 	}
@@ -84,7 +85,8 @@ func copyFile(ctx context.Context, dst, src string, disableSymlinks bool, fmode,
 
 	count, err := Copy(ctx, dstF, srcF)
 	if err != nil {
-		// Remove the file in case of partial write
+		// Close & remove the file in case of partial write
+		_ = dstF.Close()
 		_ = os.Remove(dst)
 		return 0, err
 	}


dduzgun-security (approved):
LGTM! Thanks for the PR 🙏

get_file_copy.go
```diff
@@ -47,6 +47,9 @@ func copyReader(dst string, src io.Reader, fmode, umask os.FileMode, fileSizeLim
 
 	_, err = io.Copy(dstF, src)
 	if err != nil {
+		// Close & remove the file in case of partial write
+		_ = dstF.Close()
```
jrasell, line 51: We already call close in a deferred function on line 42, so do we need this? In any case, having a second call won't cause any errors as we are throwing away the returned error.

get_file_copy.go
```diff
@@ -74,14 +77,17 @@ func copyFile(ctx context.Context, dst, src string, disableSymlinks bool, fmode,
...
+	dstF, err := os.OpenFile(dst, os.O_RDWR|os.O_CREATE|os.O_TRUNC, mode(fmode, umask))
 	if err != nil {
 		return 0, err
 	}
 	defer func() { _ = dstF.Close() }()
 
 	count, err := Copy(ctx, dstF, srcF)
 	if err != nil {
+		// Close & remove the file in case of partial write
+		_ = dstF.Close()
```
jrasell, line 89: We already call close in a deferred function on line 84, so do we need this? In any case, having a second call won't cause any errors as we are throwing away the returned error.

get_file_copy.go
allisonlarson: The windows test case would fail here because the file was reported as being in use at the time of removal. Adding in the explicit file closed fixed it and allowed the file to be removed, so it seems like windows has a different requirement/handling that needs the file to be explicitly closed before removing it. I figured it's better to close the file an extra time as opposed to trying to check.

get_file_copy.go
allisonlarson: Same reasoning as above, keeping the way these are handled in sync.

PR #634, opened 2026-06-01
remove both `.git` files and directories

Ayomi-gh:
Fixes an issue where a `.git` file containing malicious git configurations could be remain after cleanup .

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

### DIFF
--- a/get_git.go
+++ b/get_git.go
@@ -431,17 +431,18 @@ func checkGitVersion(ctx context.Context, min string) error {
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


jrasell (commented):
Hi @Ayomi-gh and thanks for this PR. Is it possible to add a test to cover this new scenario, so we can catch and avoid future regressions in this area? Thanks.

jrasell (approved):
LGTM, thanks @Ayomi-gh!

Linter error is fixed on main.

PR #635, opened 2026-06-01
ci: Enable modernize via golangci-lint. 

jrasell:


pkazmierczak (dismissed):

.github/workflows/go-getter.yml
```diff
@@ -181,7 +181,7 @@ jobs:
         uses: golangci/golangci-lint-action@1e7e51e771db61008b38414a730f564565cf7c20 # v9.2.0
 
       - name: Check for modern Go patterns
-        run: go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@latest -test ./...
+        run: go run golang.org/x/tools/gopls/internal/analysis/modernize/cmd/modernize@v0.21.1 -test ./...
```
tgross: golangci-lint supports modernize: https://golangci-lint.run/docs/linters/configuration/#modernize just not as a default. We should probably just update the lint config to pick this up so we don't have to maintain the separate depenedency.

tgross (approved):
LGTM

PR #636, opened 2026-06-01
build: Update Go to 1.25.10 and golang.org/x libs.

jrasell:
The version of Golang used for building has been updated to 1.25.10 which includes a number of security fixes. A number of golang.org/x libraries are also updated to resolve security scan failures.

The 1.26 Golang version used for additional testing in CI has also been updated to 1.26.3 which is the latest version and includes security fixes compared to 1.26.1.


.github/workflows/go-getter.yml
```diff
@@ -17,8 +17,8 @@ jobs:
     strategy:
       matrix:
         go-version:
-          - "1.25.8"
-          - "1.26.1"
+          - "1.25.10"
+          - "1.26.3"
```
tgross: In the Raft repo we use the sentinel versions `oldstable` and `stable`, which the setup-go action will use to keep this up to date for us.

tgross (dismissed):

tgross (approved):
LGTM!

PR #643, opened 2026-06-24
fix: escape bare '%' in file paths so file:// detection doesn't fail (#607)

vjsai:
Closes #607.

## Problem

`FileDetector` converts a local filesystem path into a `file://` URL by plain string concatenation (`fmtFileURL`). If the path contains a literal `%` that isn't a valid percent-encoding — e.g. a [Copier](https://copier.readthedocs.io/)/Jinja2 template directory like `{% if foo %}bar{% endif %}` — the resulting URL fails to parse downstream:

```
parse "/tmp/{% if foo %}bar{% endif %}/app": invalid URL escape "% i"
```

even though the path is perfectly valid on disk. (Reported via KCL, which uses go-getter to resolve source paths.)

## Fix

Escape only **bare** `%` characters (a `%` not followed by two hex digits) to `%25` when building the `file://` URL. The URL then parses and round-trips back to the original path via `u.Path`.

Deliberately narrow:
- valid existing escapes (`%XX`) are left as-is;
- other URL metacharacters that go-getter intentionally supports in file paths — notably a `?query` suffix (`./foo?foo=bar`) — are untouched;
- it's a **no-op for any path that already worked** (paths with no bare `%`), so existing behaviour is unchanged.

## Tests

Added `TestFileDetector_percentInPath`: detects `{% if foo %}bar{% endif %}`, parses the resulting URL, and asserts it round-trips back to the original path. It fails before this change (`invalid URL escape "% i"`) and passes after.

```
$ go test -run 'TestDetect$|TestFileDetector|TestFileGetter|TestSourceDirSubdir|TestGetFile' .
ok  	github.com/hashicorp/go-getter
$ go test ./helper/url/      # ok
$ gofmt -l ... ; go vet .    # clean
```

(The repo's GCS/S3 getter tests fail locally only because they need cloud credentials — unrelated to this change.)

### DIFF
--- a/detect_file.go
+++ b/detect_file.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 )
 
 // FileDetector implements Detector to detect file paths.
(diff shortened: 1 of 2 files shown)


tgross:
AI slop "contributor" who's been spamming repos with fixes and not signing the CLA. Closing.

PR #644, opened 2026-06-24
fix: escape bare '%' in file paths so file:// detection doesn't fail (#607)

vjsai:
Closes #607. FileDetector builds a file:// URL by string concatenation; a literal '%' in a path (e.g. a Copier/Jinja2 template dir `{% if foo %}bar{% endif %}`) then fails to parse with `invalid URL escape`. This escapes only bare '%' (not followed by two hex digits) to %25 so the URL round-trips back to the original path. Valid %XX escapes and the supported `?query` suffix are untouched; it's a no-op for paths that already worked. Adds TestFileDetector_percentInPath (fails before, passes after); local detect/file/url tests + gofmt + vet clean.

### DIFF
--- a/detect_file.go
+++ b/detect_file.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strings"
 )
 
 // FileDetector implements Detector to detect file paths.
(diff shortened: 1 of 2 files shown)


crw:
Please see https://github.com/hashicorp/go-getter/pull/643#issuecomment-4792810843

PR #663, opened 2026-07-21
feat(s3): migrate to AWS SDK v2

hariom-hashicorp:
### Summary

This PR replaces the deprecated AWS SDK for Go v1 (`github.com/aws/aws-sdk-go`) with AWS SDK for Go v2 (`github.com/aws/aws-sdk-go-v2`) across the `s3` sub-module and the `cmd/go-getter` binary. It also fixes several bugs uncovered during the migration.

---

### What changed

**SDK migration (`s3/get_s3.go`, `s3/go.mod`)**
- Replaced `session.NewSession` with `config.LoadDefaultConfig(ctx, ...)`
- Replaced `s3.New(sess)` with `s3.NewFromConfig(cfg, optFns...)`
- Replaced `s3.ListObjectsPages` callback with `s3.NewListObjectsV2Paginator`
- Replaced `awserr.Error` type assertions with `errors.As(err, &awshttp.ResponseError{})`
- Removed `jmespath/go-jmespath` transitive dependency (not used by v2)
- Bumped minimum Go version from 1.19 → 1.21 (required by `smithy-go`)

**Bug fixes (`s3/get_s3.go`)**
- `version` query parameter was silently discarded in directory downloads (`Get`) — now correctly passed to every `getObject` call
- `resp.Body` was never closed in `getObject` — added `defer resp.Body.Close()`
- Region was not applied when `aws_profile` query param is used — added `config.WithRegion(region)` to the profile-based config load

**Code cleanup (`s3/get_s3.go`)**
- Flipped condition order in `getAWSConfig`: check `staticCreds != nil` first, then fall through to IMDS metadata URL override — more explicit and direct
- Hoisted single `var err` before the `if/else` in `newS3Client` to remove duplicate declarations in each branch

**S3 URL parsing (`s3/get_s3.go`)**
- Simplified `parseUrl` to support path-style URLs only (`s3[-region].amazonaws.com/bucket/key`). An incomplete virtual-hosted-style branch was removed as dead code.

**CI (`.github/workflows/go-getter.yml`)**
- Added matrix `exclude` rules so the `s3` module is skipped on Go 1.19 and 1.20, which are incompatible with the `aws-sdk-go-v2`/`smithy-go` dependency chain.

---

### Impact & compatibility

- **No public API changes.** The `Getter` struct and all interface methods (`Mode`, `Get`, `GetFile`) are unchanged.
- **Error type change.** Code that type-asserts `awserr.Error` from S3 errors must switch to `errors.As(err, &awshttp.ResponseError{})`. Callers that only check `err != nil` are unaffected.
- **Go 1.21+ required** for the `s3` sub-module and `cmd/go-getter`. The root `go-getter/v2` module and `gcs` sub-module are not affected.

---

### Testing

- All existing unit tests pass without logic changes.
- `TestGetter_GetFile_badParams` updated to assert `*awshttp.ResponseError` (HTTP 403).
- Integration tests require AWS credentials and run in GitHub Actions CI against the `hc-go-getter-test` bucket.

**`TestGetter_Url` — credential assertion fix**

The test table previously had a single `if &creds == nil` check at the end of every sub-test. Taking the address of a local variable is never nil, so this assertion was dead code — it never actually ran regardless of what `parseUrl` returned.

When fixed to `if creds == nil`, the `AWSv1234` sub-test immediately failed because that URL has no `aws_access_key_id` / `aws_access_key_secret` query params — `parseUrl` correctly returns `nil` creds for it (signalling "use the default credential chain"). The assertion was wrong, not the code.

The fix adds a `wantCreds bool` field to the test struct so each case explicitly declares its expectation:
- `AWSv1234` → `wantCreds: false` — plain AWS URL, no inline credentials, relies on environment/IAM/config credential chain
- `localhost-1/2/3` → `wantCreds: true` — all supply `aws_access_key_id` + `aws_access_key_secret` in the URL


s3/go.mod
```diff
@@ -1,22 +1,38 @@
 module github.com/hashicorp/go-getter/s3/v2
 
-go 1.19
+go 1.21
```
tanmay-hc: Why aren't we using a more latest go version? Can we use the 1.25.* to make it more common with the `main` branch?

This applies to both the go.mod files

s3/get_s3.go
```diff
@@ -203,39 +199,36 @@ func (g *Getter) getObject(ctx context.Context, client *s3.S3, req *getter.Reque
...
-				},
+	if staticCreds == nil && metadataURLOverride != "" {
+		creds = ec2rolecreds.New(func(o *ec2rolecreds.Options) {
+			o.Client = imds.New(imds.Options{
+				Endpoint:          metadataURLOverride,
+				ClientEnableState: imds.ClientEnabled,
 			})
+		})
+	} else if staticCreds != nil {
+		creds = staticCreds
```
tanmay-hc: Should be written as:
```
if staticCreds != nil {
...
} else if metadataURLOverride != "" {
...
}

File: s3/get_s3.go [hunk 3]
```diff
@@ -338,23 +332,38 @@ func (g *Getter) validScheme(scheme string) bool {
...
 			return nil, err
 		}
 	} else {
-		config := g.getAWSConfig(region, url, creds)
-		sess = session.New(config)
+		var err error
+		cfg, err = g.getAWSConfig(ctx, region, url, creds)
+		if err != nil {
+			return nil, err
+		}
```
tanmay-hc: Can use a single err check outside if-else

s3/get_s3_test.go
tanmay-hc: Unrelated to the PR, but lets fix this
```
if creds == nil {
```

s3/go.mod
hariom-hashicorp: Thanks for the suggestion. The `main` branch currently uses `go 1.19` — bumping to `1.25.x` here would actually put this sub-module *ahead* of the rest of the repo rather than aligned with it. We settled on `1.21` because that is the lowest version supported by `smithy-go` (a hard transitive dependency of `aws-sdk-go-v2`), so it is the minimum we can honestly declare. If the project decides to raise the floor across all modules to `1.25`, happy to include that as a separate follow-up PR so it is a deliberate, repo-wide decision rather than a side effect of this migration.

tanmay-hc: The main branch currently DOES NOT use go 1.19
https://github.com/hashicorp/go-getter/blob/36dbab61b283a9407ae0533a1f9fa163559e57d1/go.mod#L84

tanmay-hc (approved):

.github/workflows/go-getter.yml
```diff
@@ -31,6 +31,12 @@ jobs:
           - s3
           - gcs
           - .
+        exclude:
+          # s3 requires Go 1.21+ (aws-sdk-go-v2 / smithy-go dependency)
+          - go-version: "1.19"
+            directory: s3
+          - go-version: "1.20"
+            directory: s3
```
tgross: Rather than excluding these, let's fix the matrix so that it's actually running against `oldstable` and `stable`, which are the only versions anyone cares about.

hariom-hashicorp: Replaced the hardcoded 1.19/1.20/1.21 version list with stable and oldstable aliases across all three jobs (linux-tests, windows-tests, go-smb-tests). The exclude blocks are no longer needed since both aliases resolve well above the go 1.25.12 floor. Also updated the Run Go cmd guard from matrix.go-version == '1.21' to matrix.go-version == 'stable'.

tgross (approved):
LGTM

PR #666, opened 2026-07-27
git: include underlying errors when resolving checkout refs

### LINKED ISSUE #649: resolveCheckoutRef silently discards underlying git errors, replacing them with a misleading "invalid ref" message
### Description

When Terraform fails to download a registry module, it can report `invalid ref: "<sha>"` — which looks like a URL or ref formatting problem — when the real cause is something completely different (e.g., a `safe.directory` ownership check, a network failure, or a permission error). The actual git error message is captured and thrown away.

### Root cause

`resolveCheckoutRef` in `get_git.go` uses `cmd.Output()` to run `git rev-parse`, which captures both stdout and stderr. When the command fails, the error output from git is silently discarded and a generic message is returned instead:

```go
for _, candidate := range candidates {
    cmd := exec.CommandContext(ctx, "git", "rev-parse", ...)
    cmd.Dir = dst

    resolvedRef, err := cmd.Output() // stderr captured here and never surfaced
    if err == nil {
        return strings.TrimSpace(string(resolvedRef)), nil
    }
    // actual git error is dropped
}

return "", fmt.Errorf("invalid ref: %q", ref) // real cause is lost
```

### Reproduction

1. Use Terraform in a dev container (or any environment) where the workspace is mounted from a host with a different user ID — a common setup on macOS/Windows.
2. Run `terraform init`.
3. Observe: `invalid ref: "<sha>"` is reported for every module.
4. The actual git error (`fatal: detected dubious ownership in repository at '...'`) is never shown.

The misleading error sends users to investigate the git ref or registry URL rather than the actual problem.

### Expected behaviour

The underlying git error(s) should be included in the returned error so the caller (and ultimately the user) can see what actually went wrong. For example:

```go
var errMsgs []string
for _, candidate := range candidates {
    cmd := exec.CommandContext(ctx, "git", "rev-parse", ...)
    cmd.Dir = dst

    resolvedRef, err := cmd.Output()
    if err == nil {
        return strings.TrimSpace(string(resolvedRef)), nil
    }
    if exitErr, ok := err.(*exec.ExitError); ok {
        errMsgs = append(errMsgs, strings.TrimSpace(string(exitErr.Stderr)))
    }
}

return "", fmt.Errorf("invalid ref %q: %s", ref, strings.Join(errMsgs, "; "))
```

### Environment

- go-getter v1.8.6 (embedded in Terraform v1.15.7)
- git 2.51.0
- Linux arm64 (dev container with macOS host volume mount)

### Additional context

Prior to the introduction of `resolveCheckoutRef`, git errors were more likely to propagate to the user. The new multi-candidate resolution approach is a useful feature, but the current error handling makes failures very difficult to diagnose.

NOTE:  The details in this issue were composed by Copilot after it diagnosed the issue.

Hashim1999164:
## Summary
- When `resolveCheckoutRef` fails every `git rev-parse` candidate, include the captured git stderr in the returned error.
- Deduplicate repeated messages across candidates so the error stays readable.
- Add a regression test that checks a non-repository directory surfaces the underlying git failure instead of only `invalid ref`.

Fixes #649

## Test plan
- [x] `go test -count=1 -run "TestGitGetter_checkoutRef|TestGitGetter_resolveCheckoutRef" .`
- [ ] Confirm a dubious-ownership / non-repo failure now includes the git message after `invalid ref`

### DIFF
--- a/get_git.go
+++ b/get_git.go
@@ -191,17 +191,46 @@ func resolveCheckoutRef(ctx context.Context, dst, ref string) (string, error) {
 		"refs/tags/" + ref,
 	}
 
+	var gitErrs []string
 	for _, candidate := range candidates {
 		cmd := exec.CommandContext(ctx, "git", "rev-parse", "--verify", "--quiet", "--end-of-options", candidate+"^{commit}")
 		cmd.Dir = dst
+
+		// Capture stderr so callers see the real git failure (for example
+		// dubious ownership) instead of only a generic invalid-ref message.
+		var stderr bytes.Buffer
+		cmd.Stderr = &stderr
 
 		resolvedRef, err := cmd.Output()
 		if err == nil {
 			return strings.TrimSpace(string(resolvedRef)), nil
 		}
+
+		msg := strings.TrimSpace(stderr.String())
+		if msg == "" {
+			msg = err.Error()
+		}
+		if msg != "" {
+			gitErrs = append(gitErrs, msg)
+		}
 	}
 
-	return "", fmt.Errorf("invalid ref: %q", ref)
+	if len(gitErrs) == 0 {
+		return "", fmt.Errorf("invalid ref: %q", ref)
+	}
+
+	// Deduplicate identical messages across candidates.
+	seen := make(map[string]struct{}, len(gitErrs))
+	uniq := make([]string, 0, len(gitErrs))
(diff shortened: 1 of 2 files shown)


Hashim1999164:
recheck

PR #673, opened 2026-08-13
Add a -version flag to the cli.

tristanmorgan:
Reason: adding a version flag to the cli so that it is identifiable.

## PCI review checklist

- [X] I have documented a clear reason for, and description of, the change I am making.

### DIFF
--- a/cmd/go-getter/main.go
+++ b/cmd/go-getter/main.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2015, 2025
+// Copyright IBM Corp. 2015, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package main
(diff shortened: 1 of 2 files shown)


cmd/go-getter/main.go
```diff
@@ -6,19 +6,26 @@ package main
...
 )
 
 func main() {
 	modeRaw := flag.String("mode", "any", "get mode (any, file, dir)")
 	progress := flag.Bool("progress", false, "display terminal progress")
 	insecure := flag.Bool("insecure", false, "do not verify server's certificate chain (not recommended)")
+	version := flag.Bool("version", false, "display version")
 	flag.Parse()
+	if *version {
+		fmt.Printf("Go-Getter v%s\n", ver.Version)
```
tgross: Shouldn't this include the other metadata in the version string? This will show 1.8.8 and main as the same.

tgross (changes requested):

cmd/go-getter/main.go
tgross: Need to update on change

```suggestion
// Copyright IBM Corp. 2015, 2026
```

cmd/go-getter/main.go
tristanmorgan: I've added a simple way of printing the extra strings but if there's an example of a better way, I would be grateful.

cmd/go-getter/main.go
```diff
@@ -1,24 +1,31 @@
...
 func main() {
 	modeRaw := flag.String("mode", "any", "get mode (any, file, dir)")
 	progress := flag.Bool("progress", false, "display terminal progress")
 	insecure := flag.Bool("insecure", false, "do not verify server's certificate chain (not recommended)")
+	version := flag.Bool("version", false, "display version")
 	flag.Parse()
+	if *version {
+		fmt.Printf("Go-Getter v%s %s %s\n", ver.Version, ver.VersionPrerelease, ver.VersionMetadata)
+		os.Exit(0)
+	}
```
tgross, line 28: Sorry, I should've been more explicit here. Typically with HashiCorp products we conditionally add the prerelease with a `-` and metadata with a `+`. Something like https://github.com/hashicorp/nomad/blob/main/version/version.go#L97-L118

tristanmorgan: I've adapted the code from Nomad and added a little formatted version string function. I felt it was cleaner than adding this to main.go

tgross (approved):
LGTM!

